PR #12, opened 2025-03-07
Move TANIP-1 issuance contracts to TANIP-1 repo

robriks:
Moves code related to TANIP-1 issuance to the [TANIP-1 repo](https://github.com/Telcoin-Association/telcoin-application-network-issuance)

This condenses related code into the TANIP-1 codebase as a monorepo using Foundry and Yarn+Jest so all related files can reference one another and be more easily found for open source viewing


grantkee (approved):
LGTM!

PR #29, opened 2025-05-02
Improve deployments json organization

robriks:
Move RWTEL, impl, manager into ITS object of deployments json. This improves upstream logic in the TN rust protocol

PR is good to go unless we want to rename RWTEL to InterchainTEL, in which case this would be a good PR to do it in


grantkee (changes requested):
Everything LGTM, but I really like the idea of updating RWTEL to InterchainTEL. Great idea. Let's change that in this PR and I will approve

grantkee (approved):

Issue #21, opened 2025-04-23
Add a new storage ledger for the outstanding stake amount


robriks:
StakingManager review included a suggestion to add a new storage ledger storing the outstanding stake amount which is more intuitive and allows incentives application to be directly applied to the outstanding stake amount.

Tradeoffs:
This would make staking/claiming more expensive but not by much. Since the ledger would only be updated en masse during system call there wouldn't be a DOS risk.

Note:
The calculation for each recipient's issuance share should still use the original stake amount which is available using the validator's stored stake version. This prevents well-capitalized entities from reaching compounding escape velocity but instead are incentivized to run more validators.

robriks:
Completed in PR #30

PR #30, opened 2025-05-03
Convert reward ledger to stake balance ledger

robriks:
- Renames `IncentiveInfo` to `StakeInfo` and introduces slashing-specific `Slash` struct instead of reusing IncentiveInfo
- Converts rewards ledger `IncentiveInfo.stakingRewards` to outstanding balance ledger `StakeInfo.balance`

Previously the ConsensusRegistry did not store stake amounts by ensuring contract balance could not be altered outside of staking/unstaking. The incentiveInfo mapping served as a ledger only for rewards, which were wiped on each claim to prevent reentrancy. While minimalistic, that approach was very rigid for future slashing implementations and generally unintuitive. The ledger now stores each validator's full outstanding balance and prevents the introduced reentrancy possibility with ReentrancyGuard

- Updates rewards checks & distribution to handle new ledger
- Moves functions to more appropriate groupings and rename where appropriate

Stake delegation fns `delegationDigest` `_getRecipient` `_checkKnownDelegation` -> `StakeManager`
`_checkRewardsExceedMinWithdrawAmount` -> `_checkRewards` to reflect new ledger checks
`ConsensusRegistry::_checkKnownValidator` -> `StakeManager::_checkConsensusNFTOwner`

- Slash future-proofing: consolidates stake burned by slashes in the InterchainTEL contract for future optionality

Completes #21


robriks:
After some discussion on epoch issuance rewards, preferences converged on a manual approach for issuing consensus block rewards. 

Because TEL is currently a fixed supply token, issuance cannot be made via inflation without significant community feedback and the governance process. The inflation option can be explored at a later time, but for MNO pilot the consensus block rewards must come from an existing TEL source, for which the TAO has generously approved budget allocation.

A new `Issuance.sol` contract was designed to accept TEL for this purpose: the TAO governance safe alone can burn TEL allocated for genesis block rewards by sending to the Issuance contract via `StakeManager::allocateIssuance()`

A separate contract is used because the ConsensusRegistry's TEL balance would be agnostically decremented when distributing rewards since the `ConsensusRegistry::StakeInfo.balance` ledger now includes both stake and rewards. This would lead to distributed rewards effectively being pulled from other validators' balances if TAO governance failed to allocate. Instead, the separate Issuance contract offloads accounting to the EVM native ledger.

Staker claim flow is pull-only, since `ConsensusRegistry::applyIncentives()` only increments the rewards owed and the only methods for stakers to claim rewards are `ConsensusRegistry::claimStakeRewards()` or a full retirement (stake + rewards) using `ConsensusRegistry::unstake()`. Thus `applyIncentives()` will never revert a critical protocol system call or epoch action. 

Claims will revert in the case that TAO governance fails to allocate, but in that scenario the ledger still serves as a useful IOU which informs exactly how much consensus block rewards are owed, using `ConsensusRegistry::getRewards()`

`applyIncentives` and `applySlashes` were separated out from `concludeEpoch` so that they can be enabled at will later on sometime after MNO testnet pilot. If enabled, they must be system called *just before* epoch conclusion

grantkee (approved):

PR #31, opened 2025-05-04
Offload decimal handling to Axelar Hub

robriks:
Given TN native gas currency needs and the pre-existing TEL token's implementation, the contract-layer ITS protocol requires TN to designate an ERC20 contract which inherits `InterchainTokenStandard` and assign it an interchain token ID. This is the `InterchainTEL` contract which for maximum flexibility makes use of the `custom-linked` ITS token ID schema.

One major feature of `custom-linked` ITS token IDs is the ability to link pre-existing tokens across chains, which requires translation of ERC20 metadata `name || symbol || decimals` when they differ between chains or are deployed on alt-VM chains. This PR offloads said decimal handling to the Axelar Hub.

The ideal flow for enabling `InterchainTEL` to flow from Ethereum to TN at genesis using ITS protocol as precompiles is as follows:

1. Declare TEL on Ethereum as an ITS custom-linked token: `InterchainTokenService::registerTokenMetadata()`
- This instructs the Axelar Hub that `InterchainTEL` is ready to be linked

2. Perform the mirror action for TN ahead of genesis (before TN exists!). There will not yet be a service contract with `registerTokenMetadata()` yet, so declaring `InterchainTEL` on Telcoin Network as the corresponding ITS custom-linked token ID can be done in either of 2 ways:

  - by pre-signing the corresponding `InterchainTokenService::MESSAGE_TYPE_REGISTER_TOKEN_METADATA` message on TelcoinNetwork using TN verifier keys and feeding them into the TN voting verifier before TN genesis. no forking necessary
  - by forking the voting verifier on Axelar Network so that the metadata message can bypass verification a single time (must not be replayable). theoretically could be frontrun but highly doubt it would be

3. Once metadata has been registered telling Axelar Hub that `InterchainTEL` is ready to be linked and has a decimal diff (even though the network hasn't begun yet), perform the TEL link to TN on Ethereum pre-genesis using `InterchainTokenFactory::linkToken()`

(bonus # 4. The initial bridge transaction into TN can be a genesis system call delivering 0x7e1 to a relayer who then can begin processing more bridge txs. That 0x7e1 would need to be burned on Ethereum)

Hub decimal logic:

https://github.com/axelarnetwork/axelar-amplifier/blob/aa956eed0bb48b3b14d20fdc6b93deb129c02bea/contracts/interchain-token-service/src/contract/execute/mod.rs#L260

https://github.com/axelarnetwork/axelar-amplifier/blob/aa956eed0bb48b3b14d20fdc6b93deb129c02bea/contracts/interchain-token-service/src/contract/execute/interceptors.rs#L228


grantkee (approved):
LGTM! I'm in favor of approach 2a, but I think both are technically feasible

PR #37, opened 2025-06-02
Add Telcoin SVG as StakeManager::tokenURI

robriks:
Adds Telcoin SVG as read-only image URI for all tokenIds


grantkee (approved):
Sorry this took so long to review. LGTM!

PR #40, opened 2025-06-12
Live GMP & ITS messages

robriks:
Implement & document GMP + ITS messages
GMP is live between Sepolia & local net
ITS requires a whitelisting step on the Axelar `eth-sepolia` gateway but is currently sidestepped in simulations

Token linking requires multiple transactions before any ITS messages can be routed to Telcoin Network from Sepolia/Ethereum
1. `originChain::ITF::registerCustomToken()`
2. `originChain::ITF::linkToken()`
3. `remoteChain::ITS::registerTokenMetadata()`
4. `remoteChain::gateway::approve(MESSAGE_TYPE_LINK_TOKEN)`
5. `remoteChain::gateway::execute(MESSAGE_TYPE_LINK_TOKEN)`

Then the first bridge transactions can come through. Assuming they have been lined up ahead of time, only two transactions remain:
6. `remoteChain::gateway::approve(MESSAGE_TYPE_INTERCHAIN_TRANSFER)`
7. `remoteChain::gateway::execute(MESSAGE_TYPE_INTERCHAIN_TRANSFER)`

Events in the first 3 transactions must be queriable via RPC within a "final" block. The RPC could probably fudge truthiness for those queries. Alternatively, the linking steps can be handled at the Axelar Network + verifier level by forking the voting verifier for expedited linking or by pre-obtaining signatures for each event query. Another option is to bypass the need for all of this to happen at genesis by pre-burning and allocating funds to an address on Telcoin-Network which can then perform all of the above actions normally on a live network.

### DIFF
--- a/deployments/utils/ITSConfig.sol
+++ b/deployments/utils/ITSConfig.sol
@@ -31,6 +31,7 @@ abstract contract ITSConfig is ITSUtils {
     uint256 constant SEPOLIA_CHAINID = 11155111;
     string constant DEVNET_TN_CHAIN_NAME = "telcoin";
     bytes32 constant DEVNET_TN_CHAINNAMEHASH = keccak256(bytes(DEVNET_TN_CHAIN_NAME));
+    bytes32 constant DEVNET_INTERCHAIN_TOKENID = 0x09f3c6e8754c8e269060f138d8b55526b0058b0fa258f90ae3197ee4808d23b6;
     string constant DEVNET_SEPOLIA_CHAIN_NAME = "eth-sepolia";
     bytes32 constant DEVNET_SEPOLIA_CHAINNAMEHASH = 0x24f78f6b35533491ef3d467d5e8306033cca94049b9b76db747dfc786df43f86;
     address constant DEVNET_SEPOLIA_ITS = 0x2269B93c8D8D4AfcE9786d2940F5Fcd4386Db7ff;
--- a/script/InterchainTransfer.s.sol
+++ b/script/InterchainTransfer.s.sol
@@ -12,15 +12,14 @@ import { InterchainTEL } from "../src/InterchainTEL.sol";
 import { Deployments, ITS } from "../deployments/Deployments.sol";
 import { ITSConfig } from "../deployments/utils/ITSConfig.sol";
 
-/// @dev Usage: `forge script script/LinkToken.s.sol -vvvv`
-contract LinkToken is ITSConfig, Script {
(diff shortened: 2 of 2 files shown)


node/src/relay/devnet/relay-instructions.md
grantkee, line 1: nice

node/src/relay/utils.ts
```diff
@@ -24,21 +24,38 @@ export interface TargetConfig {
...
+          break;
+        case "ethereum-sepolia": // testnet
           targetConfig.chain = sepolia;
           setRpcUrl("SEPOLIA_RPC_URL");
           break;
         case "ethereum":
           targetConfig.chain = mainnet;
           setRpcUrl("MAINNET_RPC_URL");
           break;
+        case "telcoin": // devnet
```
grantkee, line 55: nit: "local" might be better since we plan to support a dev, test, and mainnet

node/src/relay/utils.ts
```diff
@@ -103,6 +202,86 @@ export function createHttpsAgent(
...
+    "auto",
+    "--gas-adjustment",
+    "1.5",
+  ];
+  console.log(`Running bash: \naxelard ${axelardArgs.join(" \\\n  ")}`);
+
+  const axelardProcess = spawn("axelard", axelardArgs);
+
+  if (process.env.PASSPHRASE) {
+    axelardProcess.stdin.write(`${process.env.PASSPHRASE}\n`);
```
grantkee, line 238: Probably best to close `stdin` here since there's no more expected input
`axelardProcess.stdin.end();`

node/src/relay/utils.ts
```diff
@@ -103,6 +202,86 @@ export function createHttpsAgent(
...
+    console.error("Must set PASSPHRASE in .env");
+    axelardProcess.kill();
+    return;
+  }
+
+  let stdoutData = "";
+
+  axelardProcess.stdout.on("data", (chunk) => {
+    stdoutData += chunk.toString();
+  });
```
grantkee, line 249: I think you can collect `Buffer`s here which might be more performant, but this works so we can revisit this later
```
    const stdoutChunks: Buffer[] = [];

    axelardProcess.stdout.on("data", (chunk: Buffer) => {
      stdoutChunks.push(chunk);
    });

    // later
     const stdoutData = Buffer.concat(stdoutChunks).toString();
```

We can discuss this as an optimization, but this PR is good to go without refactoring

grantkee (approved):
LGTM! I had a few comments, but nothing is a show stopper

PR #41, opened 2025-06-13
Add Safe smart account infra as precompiles and governance safe as base fee recipient

robriks:
Adds three precompiles to the genesis config yaml:
- Safe singleton implementation
- SafeProxyFactory
- One Safe proxy, to serve as governance address

The governance safe's role is manifold:
1. Base fee recipient on a per-epoch basis
2. Recipient for truncated TEL during outbound ITS bridge txs
3. Owner for `InterchainTEL` and `ConsensusRegistry`
4. Initial ITS relayer for Telcoin Network, allocated a small initial TEL balance during genesis for gas which allows inbound ITS bridge execution, delivering inbound TEL for further relaying

The governance safe's address is `0x00000000000000000000000000000000000007a0` and is currently allocated 10 TEL at genesis.


grantkee (approved):
LGTM! Good work on this

PR #55, opened 2025-09-19
Add third chain demonstration to clarify Axelar options

robriks:
Adds a third chain to the custom token linking fork test, demonstrating that Axelar's custom linked tokenID schema supports multiple instances of either MINT_BURN && LOCK_RELEASE token managers. This is in direct contradiction to their documentation, which appears to be outdated and based only on the canonical token ID schema.

Based on discussions, use of `MINT_BURN` token manager with a "newTEL" token upgraded from "oldTEL" in place on Base & Polygon is more desirable.

Link test notes:
- Axelar amplifier devnet does not support Base or Polygon, so "optimism-sepolia" was chosen as the closest standin. 
- The fork test for linking has also been extended to a fuzz test to fuzz between both token manager types.

### DIFF
--- a/deployments/Deployments.sol
+++ b/deployments/Deployments.sol
@@ -21,6 +21,7 @@ struct Deployments {
     address admin;
     EXYZs eXYZs;
     ITS its;
+    address optimismTEL;
     address sepoliaTEL;
     UniswapV2 uniswapV2;
     address wTEL;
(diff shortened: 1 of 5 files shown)


grantkee (approved):
Thank you for updating these tests. Great work!

PR #57, opened 2025-10-09
Precompiles: Historic Block Hashes & Consensus Roots

grantkee:
add bytecode to its-config.yaml for deploying historic blockhashes and consensus root hashes in genesis for cancun fork


robriks:
This looks great and is technically good to merge on the Solidity end. 

There is one interesting possibility introduced here where the `create` transaction for these contracts for EIP 2935 and 4788 (`0x0000F90827F1C53a10cb7A02335B175320002935` and `0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02` respectively) can be replayed since the synthetic EOA senders for both create transactions still has a nonce of 0 at genesis.

As a result, the raw RLP-encoded transaction which deploys the _correct_ bytecode to the two contract addresses named above can still be published and executed, overwriting the genesis configuration set in this PR. However that should have no effect on anything beyond duplicating existing state and wasting gas. It is exceedingly unlikely, probabistically impossible really, that a transaction can be built to recover from the synthetic EOA sender addresses (`0x3462413Af4609098e1E27A490f554f260213D685` and `0x0B799C86a49DEeb90402691F1041aa3AF2d3C875` respectively) which could deploy arbitrary bytecode to the two EIP contracts. So it is technically safe, but for completeness I think we should increment the nonces of both synthetic senders anyway so there is no possibility of weird duplicate overwrites.

What do you think @grantkee ?

robriks (approved):

Issue #74, opened 2026-03-04
Byte array size mismatch when comparing `uncompressedSignature` to `G1_IDENTITY`


chasebrownn:
This issue refers to a possible bug within [BlsG1::encodeG1PointForEIP2537](https://github.com/Telcoin-Association/tn-contracts/blob/84a7c4dfa50b81edcaf8d6b21007ab03a247f6fa/src/consensus/BlsG1.sol#L97). 

At line 104 we call `isInfinityPointG1`, passing our `uncompressedSignature` G1 point. This checks whether our raw G1 point is equal to our G1 Identity.

However, `uncompressedSignature` is 96 bytes (if not we'd revert at line 99) and `G1_IDENTITY` is 128 bytes so this will always return false.

PR #75, opened 2026-03-06
Pushing fix for point-identity comparison on encoding functions

chasebrownn:
Solution to issue: https://github.com/Telcoin-Association/tn-contracts/issues/74

Small issue when comparing our G1 point to the G1 Identity and G2 point to the G2 identity during the `encodeG*PointForEIP2537` executions.

The encoded identities did not carry the same length as the points it was compared to, returning false with every call (aka, dead code).

Solution just compares the points to pre-encoded identities matching the size of the G1/G2 points.

### DIFF
--- a/src/consensus/BlsG1.sol
+++ b/src/consensus/BlsG1.sol
@@ -105,7 +105,7 @@ library BlsG1 {
 
         // if the provided signature is the identity, return early with 128 zero bytes
         bytes memory encoded = new bytes(EIP2537_G1_POINT_SIZE);
-        if (bytesEq(uncompressedSignature, G1_IDENTITY_RAW)) return encoded;
+        if (isUncompressedInfinityPointG1(uncompressedSignature)) return encoded;
 
         // otherwise, pad x and y coordinates with 16 leading zeroes
         fieldElementToEIP2537Bytes(uncompressedSignature, 0, encoded, 0);
@@ -145,7 +145,7 @@ library BlsG1 {
 
         // if the provided public key is the identity, return early with 256 zero bytes
         bytes memory encoded = new bytes(EIP2537_G2_POINT_SIZE);
-        if (bytesEq(uncompressedPubkey, G2_IDENTITY_RAW)) return encoded;
+        if (isUncompressedInfinityPointG2(uncompressedPubkey)) return encoded;
 
         // reorder blst serialization of coordinates for EIP2537 compliance: `x.c0 || x.c1 || y.c0 || y.c1`
         // x.c0 (second 48 bytes of x coordinate, bytes 48-96)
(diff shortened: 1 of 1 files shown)


robriks (changes requested):
Thanks for tracking this down and submitting the PR! You're spot on about the root cause being the mismatched byte lengths in the identity comparison.

I'd prefer not to just inline the fix (as done with `bytesEq`) while still leaving the existing `isInfinityPointGx` functions untouched. Because those original functions use parameters named `uncompressedPointG*` but check against EIP-2537 padded lengths, they still carry the existing bug and are misleading to leave in. 

The library handles both uncompressed (96/192 bytes) and EIP-2537 encoded (128/256 bytes) formats depending on the function, so we actually need explicit checks for both states and can provide explicitly named functions to clearly delineate them.

To keep the library clean, could we structure the fix like this:

1. Rename the parameters in the existing EIP-2537 checks:
To clarify that these expect the 128/256-byte encoded formats, let's update their parameter names. (These will continue to be used in `validatePointG1/G2` and `verifyProofOfPossessionG1`).

```Solidity
    /// @dev Checks if an EIP2537-encoded G1 point is the identity
    function isInfinityPointG1(bytes memory eip2537PointG1) public pure returns (bool) {
        return bytesEq(eip2537PointG1, G1_IDENTITY);
    }

    /// @dev Checks if an EIP2537-encoded G2 point is the identity
    function isInfinityPointG2(bytes memory eip2537PointG2) public pure returns (bool) {
        return bytesEq(eip2537PointG2, G2_IDENTITY);
    }
```

2. Add new uncompressed check fns utilizing your `_RAW` constants:
Add these right below the existing checks to handle the 96/192-byte formats.

```Solidity
    /// @dev Checks if an uncompressed G1 point (96 bytes) is the identity
    function isUncompressedInfinityPointG1(bytes memory uncompressedPointG1) public pure returns (bool) {
        return bytesEq(uncompressedPointG1, G1_IDENTITY_RAW);
    }

    /// @dev Checks if an uncompressed G2 point (192 bytes) is the identity
    function isUncompressedInfinityPointG2(bytes memory uncompressedPointG2) public pure returns (bool) {
        return bytesEq(uncompressedPointG2, G2_IDENTITY_RAW);
    }
```

3. Update the encode functions to use the new uncompressed checks:
Instead of inlining the `bytesEq` logic, call the new helpers in `encodeG1PointForEIP2537` and `encodeG2PointForEIP2537`.

```Solidity
        // In encodeG1PointForEIP2537:
        bytes memory encoded = new bytes(EIP2537_G1_POINT_SIZE);
        if (isUncompressedInfinityPointG1(uncompressedSignature)) return encoded;
```

```Solidity
        // In encodeG2PointForEIP2537:
        bytes memory encoded = new bytes(EIP2537_G2_POINT_SIZE);
        if (isUncompressedInfinityPointG2(uncompressedPubkey)) return encoded;
```

Let me know if you have time to update the PR with this structure! This will keep the cryptographic baseline rock-solid.

PR #77, opened 2026-03-16
Remove ITS

grantkee:
- remove ITS
- remove node/npm to use foundry only

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,9 +4,6 @@
 [submodule "lib/solady"]
 	path = lib/solady
 	url = https://github.com/Vectorized/solady
-[submodule "lib/openzeppelin-contracts"]
-	path = lib/openzeppelin-contracts
-	url = https://github.com/OpenZeppelin/openzeppelin-contracts
 [submodule "lib/openzeppelin-contracts-upgradeable"]
 	path = lib/openzeppelin-contracts-upgradeable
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -0,0 +1,18 @@
+[submodule "lib/forge-std"]
+	path = lib/forge-std
+	url = https://github.com/foundry-rs/forge-std
+[submodule "lib/solady"]
+	path = lib/solady
+	url = https://github.com/Vectorized/solady
+[submodule "lib/openzeppelin-contracts"]
+	path = lib/openzeppelin-contracts
+	url = https://github.com/OpenZeppelin/openzeppelin-contracts
```
chasebrownn: nit: Something I’ve started doing in my Forge environments is mapping `@openzeppelin/contracts` to the version that already exists inside the `openzeppelin-contracts-upgradeable` submodule. Since it’s the exact same OZ version, this lets us avoid importing the base `openzeppelin-contracts` module separately and keeps the repo a bit cleaner.

The upgradeable package already references the same contracts internally, so this ends up being functionally identical while reducing the number of top-level submodules.

Concretely, you could:

forge remove lib/openzeppelin-contracts

Then update `remappings.txt` to:

@openzeppelin/contracts/=lib/openzeppelin-contracts-upgradeable/lib/openzeppelin-contracts/contracts/

Minor change overall — same contracts, just one less parent-level submodule.

PR #88, opened 2026-03-31
EpochTargets -> WorkerConfigs

### LINKED ISSUE #89: Replace EpochGasTarget with a strategy-agnostic WorkerConfigs contract
## Problem

The current `EpochGasTarget` contract stores a single `uint64` gas target per worker, with a global default that unset workers fall back to. This design implicitly assumes every worker uses the same fee strategy (EIP-1559 style gas targeting), making it impossible to assign different fee strategies to different workers without deploying additional contracts or encoding strategy information out-of-band.

As the protocol matures, workers will need to support heterogeneous fee strategies -- for example, EIP-1559 dynamic targeting on some workers and static fee schedules on others. The current contract has no concept of a strategy identifier, so the execution layer has no on-chain signal for which fee algorithm to apply to a given worker. Any workaround (off-chain config, hardcoded mappings in the node) would fragment the source of truth and create synchronization risks between the contract and the protocol.

Additionally, `EpochGasTarget` has no notion of a canonical worker count. The protocol must independently track how many workers exist, and there is no on-chain guarantee that every active worker has a valid configuration. A worker could silently fall back to the global default even when the operator intended to set a specific target, with no revert or event to flag the gap.

## Solution

Introduce a `WorkerConfigs` contract that replaces `EpochGasTarget` with a strategy-agnostic per-worker config store. Each worker should be identified by a `uint16` and mapped to a `(uint8 strategy, uint64 value)` pair, where `strategy` is a raw identifier stored without contract-level interpretation and `value` carries the fee parameter (e.g., gas target for EIP-1559, fixed fee for static strategies).

The contract should maintain an explicit `numWorkers` count so the protocol can iterate `getWorkerConfig(0 .. numWorkers-1)` at each epoch boundary to build the full fee parameter set. The constructor should atomically set both the worker count and every worker's config, guaranteeing full coverage from deployment.

`setNumWorkers` should validate that every worker in the range `0 .. numWorkers-1` has a config with `value >= MIN_GAS` (7 wei, matching `MIN_PROTOCOL_BASE_FEE` on the Rust side), reverting with a descriptive error if any worker is unconfigured. `setWorkerConfig` should allow pre-configuring workers beyond the current count so governance can stage configs before expanding the worker set (ideally batched via multicall).

This design keeps strategy semantics entirely in the protocol layer while giving governance a single on-chain source of truth for both the worker set size and each worker's fee policy.

grantkee:
## Overview

Replaces the `EpochGasTarget` contract with a strategy-agnostic `WorkerConfigs` contract that stores a `(strategy, value)` pair per worker and maintains an explicit worker count with coverage validation. This is another step towards moving network-wide configuration on-chain.

### WorkerConfigs contract (`src/consensus/WorkerConfigs.sol`)
- New contract replacing `EpochGasTarget`, storing a `WorkerConfig { uint8 strategy, uint64 value }` per worker in a `mapping(uint16 => WorkerConfig)`
- Constructor takes parallel `strategies[]` and `values[]` arrays plus owner, atomically setting `numWorkers` and every worker's config with full validation
- `setNumWorkers(uint16)` validates that every worker `0..n-1` has `value >= MIN_GAS` before allowing the count to change, reverting with `MissingWorkerConfig(workerId)` on gaps
- `setWorkerConfig(uint16, uint8, uint64)` allows configuring any workerId (including beyond current `numWorkers`) to support pre-staging before expansion
- `getWorkerConfig(uint16)` returns `(strategy, value)` for any workerId; returns `(0, 0)` for unconfigured workers
- `MIN_GAS` constant set to 7 wei, matching `MIN_PROTOCOL_BASE_FEE` on the Rust side

### IWorkerConfigs interface (`src/interfaces/IWorkerConfigs.sol`)
- Full interface with errors (`ValueBelowMinGas`, `MissingWorkerConfig`, `LengthMismatch`, `NumWorkersBelowMinimum`, `TooManyWorkers`), events (`WorkerConfigUpdated`, `NumWorkersUpdated`), and all public function signatures
- Replaces `IEpochGasTarget`

### EpochGasTarget removal
- Deleted `src/consensus/EpochGasTarget.sol`, `src/interfaces/IEpochGasTarget.sol`, and `test/consensus/EpochGasTargetTest.t.sol`
- Deleted `artifacts/EpochGasTarget.json`

### Test suite (`test/consensus/WorkerConfigsTest.t.sol`)
- 290-line test file covering constructor validation (length mismatch, zero workers, value below `MIN_GAS`, event emission), `setNumWorkers` (expansion, shrinking, coverage check failures, zero revert, idempotent case), `setWorkerConfig` (any strategy, beyond `numWorkers`, overwrites, `MIN_GAS` boundary), and `getWorkerConfig` (stored vs unset)
- Fuzz tests for arbitrary `(workerId, strategy, value)` configs and parameterized worker counts up to 20

### Invariants documentation (`src/consensus/invariants.md`)
- Added "worker configs" and "worker configs -- protocol usage" sections documenting coverage guarantees, `MIN_GAS` enforcement, stale config behavior on shrink/expand, and epoch-boundary read pattern

### Deployment and build
- `deployments/Deployments.sol`: renamed `EpochGasTarget` field to `WorkerConfigs`
- `deployments/deployments.json`: updated key from `EpochGasTarget` to `WorkerConfigs`
- `Makefile`: updated `update-artifacts` target to copy `WorkerConfigs.json` instead of `EpochGasTarget.json`
- Added `artifacts/WorkerConfigs.json` with the new ABI

closes #89 
related to [telcoin-network #610](https://github.com/Telcoin-Association/telcoin-network/issues/610)


chasebrownn (commented):
A couple small things

PR #93, opened 2026-04-03
Validator nft geographic data

grantkee:
### Core feature
- Added `uint8 region` (GSMA region identifier) to `ValidatorInfo` struct, packed into the existing storage slot
- Added `setValidatorRegion(address, uint8)` governance function with `onlyOwner` + ConsensusNFT ownership check
- Added `ValidatorRegionUpdated(address indexed, uint8)` event
- Region defaults to 0 (unspecified) on stake; full uint8 range (0-255) accepted with no max constraint

### `isDelegated` refactor
- Removed `bool isDelegated` from `ValidatorInfo` struct to prevent storage slot overflow when adding `region`
- Added `isDelegated(address) view` function that checks `delegations[addr].delegator != address(0)`
- Removed genesis-time delegation initialization from `_processGenesis` (delegation records are set during `delegateStake`, not genesis)

### Events
- Added `indexed` to `RewardsClaimed(address indexed claimant, uint256 rewards)`
- Added `indexed` to `ValidatorRegionUpdated(address indexed validatorAddress, uint8 region)`

### Removed: `EpochGasTarget`
- Deleted `EpochGasTarget.sol`, `IEpochGasTarget.sol`, and `EpochGasTargetTest.t.sol`
- Removed from `Deployments.sol` and `deployments.json`
- Removed associated artifacts

### Documentation
- Added Geographic Diversity section to `design.md` covering region mapping, `setValidatorRegion`, and shuffle algorithm
- Added region invariants to `invariants.md`
- Noted that the region-aware shuffle lives in the Rust protocol client, not in Solidity

### Tests
- Unit tests for `setValidatorRegion` across all validator statuses (Active, Staked, PendingActivation, PendingExit, Exited)
- Tests for edge cases: max uint8 (255), reset to 0, no-NFT revert, non-owner revert
- Fuzz test covering full uint8 range
- Updated all `ValidatorInfo` constructors in tests to reflect struct change (`isDelegated` -> `stakeVersion`, `region`)


src/consensus/ConsensusRegistry.sol
```diff
@@ -234,6 +241,11 @@ contract ConsensusRegistry is StakeManager, Pausable, Ownable, ReentrancyGuard,
         return true;
     }
 
+    /// @inheritdoc IConsensusRegistry
+    function isDelegated(address validatorAddress) external view returns (bool) {
+        return _isDelegated(validatorAddress);
+    }
```
chasebrownn, line 247: Do we want this to return `true` if a validator retires? Seems this state does not change if there is an unstake or burn of the validator.

src/consensus/ConsensusRegistry.sol
```diff
@@ -132,6 +132,13 @@ contract ConsensusRegistry is StakeManager, Pausable, Ownable, ReentrancyGuard,
         }
     }
 
+    /// @inheritdoc IConsensusRegistry
+    function setValidatorRegion(address validatorAddress, uint8 region) external onlyOwner {
+        _checkConsensusNFTOwner(validatorAddress);
+        validators[validatorAddress].region = region;
+        emit ValidatorRegionUpdated(validatorAddress, region);
+    }
```
chasebrownn, line 140: Should we sanitize `region` for invalid regions? I know we're only supporting 0-8 currently, but that leaves 9-255 as possible invalid inputs that could throw things off.

src/consensus/ConsensusRegistry.sol
```diff
@@ -612,21 +630,13 @@ contract ConsensusRegistry is StakeManager, Pausable, Ownable, ReentrancyGuard,
...
         ValidatorInfo memory newValidator = ValidatorInfo(
-            blsPubkey,
-            validatorAddress,
-            PENDING_EPOCH,
-            uint32(0),
-            ValidatorStatus.Staked,
-            false,
-            isDelegated,
-            stakeVersion
+            blsPubkey, validatorAddress, PENDING_EPOCH, uint32(0), ValidatorStatus.Staked, false, stakeVersion, uint8(0)
```
chasebrownn, line 639: Do we not want to copy over the existing region value to the new validator?

chasebrownn (commented):
Just a few minor thoughts - nothing crazy.

src/consensus/ConsensusRegistry.sol
grantkee: Good catch. Validator delegations were not included in the exit path. They should return `false` now after exit/burn

src/consensus/ConsensusRegistry.sol
grantkee: I'm not sure exactly how we want to use regions. GSMA uses 1-8, but we might need to be more granular with assigning validator regions. I'm intentionally leaving this open for max flexibility and will ensure the protocol is flexible to handle all variations

src/consensus/ConsensusRegistry.sol
grantkee: Each validator's region will be applied after they stake. `0` will be default and is always considered in the shuffle. I think this will give us the most flexibility moving forward so validators aren't excluded by default but governance can adjust committees by region if/when necessary

src/consensus/ConsensusRegistry.sol
Huwonk: I'd err on the side of having a definition stored/enforced.  I'd generally prefer deliberate expansion of acceptable codes, somewhat in line with Chase's comment, however I do understand the dilemma.  Non-blocker either way imo.

src/consensus/ConsensusRegistry.sol
```diff
@@ -1034,11 +1051,6 @@ contract ConsensusRegistry is StakeManager, Pausable, Ownable, ReentrancyGuard,
             if (currentValidator.isRetired != false) {
                 revert InvalidStatus(ValidatorStatus.Exited);
             }
-            if (currentValidator.isDelegated == true) {
-                // at genesis, only governance delegations are enabled
-                delegations[currentValidator.validatorAddress] =
-                    Delegation(blsPubkeyHash, currentValidator.validatorAddress, owner_, uint8(0), uint64(1));
-            }
```
Huwonk, line 1041: Trying to understand this removal.  This deleted block was the only writer to delegations[] at construction. Post-merge, no genesis validator can have a delegator, because delegateStake requires ValidatorStatus.Undefined and genesis validators are seeded with Active.  Is this intentional?  After this PR, no genesis validator can ever have a non-self delegator for the lifetime of the contract.  This may be intentional.

Huwonk (changes requested):
Looks good other than the one question I had.  If that is good, it's ready to merge imo.

src/consensus/ConsensusRegistry.sol
Huwonk: Confirmed with Grant this is ok.

Huwonk (approved):
All looks good to me.

Huwonk (approved):
master branch seemed to have no changes pushed back to this.  Only artifact changes.  Looks good.

PR #95, opened 2026-04-24
Fix native faucet drip reverting on the TEL precompile's EXTCODESIZE guard

### LINKED ISSUE #94: Native drip from `StablecoinManager` reverts on-chain due to Solidity's EXTCODESIZE guard on the TEL precompile
## Problem

`StablecoinManager.drip(address(0))` is the public entry point for claiming the native TEL faucet. Its native path calls the TEL mint precompile:

```
ITELMint public constant TEL_MINT = ITELMint(address(0x7e1));
// ...
TEL_MINT.mint(recipient, amount);
```

On the live network, address `0x07e1` has no bytecode in state — the TEL mint precompile is dispatched by revm at execution time, not deployed as an EVM contract. Solidity inserts an automatic EXTCODESIZE check ahead of every typed-interface external call. When `EXTCODESIZE(0x07e1) == 0`, that guard reverts with an empty revert reason *before* the precompile is ever reached.

The result is that every native faucet drip reverts on live chains, even though the same code path succeeds in unit tests that use `vm.etch` or `vm.mockCall` (both of which give the address non-zero code size and mask the real-chain behavior). The faucet is effectively dead for native TEL on testnet until this is resolved. ERC20 drips are unaffected because those targets are real deployed contracts with bytecode.

The project already hit and solved the same class of issue for the BLS12-381 precompiles in `BlsG1.sol`, which uses a low-level `.staticcall` pattern specifically to bypass the compiler-emitted codesize check. The faucet path was written with a typed interface instead, and the mismatch between unit-test behavior and on-chain behavior kept the regression hidden.

## Solution

`StablecoinManager` should invoke the TEL precompile the same way `BlsG1.sol` invokes the BLS precompiles: via a low-level `.call` that does not emit a codesize guard. The selector and arguments should be ABI-encoded explicitly, the return tuple captured, and a non-success return should revert with a typed error that carries the raw return data so failures remain diagnosable.

The existing `LowLevelCallFailure` error should be widened to include the returned bytes (matching the `BlsG1` convention), and the one other site in the contract that uses it — the native branch of `rescueCrypto` — should be updated to pass the return data through for consistency.

Test coverage should include a regression guard for the drip path that deliberately does **not** use `vm.etch` or `vm.mockCall` on `0x07e1`, leaving its code size at zero. This replicates real-chain conditions and ensures the low-level path cannot quietly regress to a typed-interface call under a future refactor.

grantkee:
- Replace the typed `TEL_MINT.mint(...)` call in `StablecoinManager.drip` with a low-level `.call` so the compiler-emitted EXTCODESIZE check no longer reverts on `0x07e1`, which has no bytecode on the live chain (revm dispatches the precompile at runtime). Same pattern already used in `BlsG1.sol`.
- Widen `LowLevelCallFailure` to carry the returned bytes, and propagate the return data from the native branch of `rescueCrypto` for parity.
- Add a regression test that exercises the native drip with `code.length == 0` at `0x07e1` — no `vm.etch`, no `vm.mockCall` — so the failure mode can't creep back in under unit tests that would otherwise mask it.

closes #94

### DIFF
--- a/src/testnet/StablecoinManager.sol
+++ b/src/testnet/StablecoinManager.sol
@@ -26,7 +26,7 @@ contract StablecoinManager is StablecoinHandler, TNFaucet, UUPSUpgradeable {
         uint256 decimals;
     }
 
-    error LowLevelCallFailure();
+    error LowLevelCallFailure(bytes returnData);
     error InvalidOrDisabled(address token);
     error AlreadyEnabled(address token);
     error InvalidDripAmount(uint256 dripAmount);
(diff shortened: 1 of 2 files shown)


Huwonk (approved):
Minor change, looks good to me.

PR #97, opened 2026-04-27
Separate validator BLS keys into their own mapping

### LINKED ISSUE #96: refactor: ValidatorInfo embedded BLS pubkey
## Problem

`IConsensusRegistry.ValidatorInfo` carries a 256-byte `bytes blsPubkey` field next to fixed-width metadata: validator address, two epoch numbers, status enum, two booleans, and a stake version. Solidity stores a dynamic `bytes` field inside a storage struct as a length slot at the struct's position plus a separately-hashed payload region. That layout forces every validator entry to occupy at least two hot slots before a single byte of pubkey is read, and the payload itself spans eight more slots for an EIP-2537 uncompressed key.

The remaining fixed-width fields fit in 32 bytes. With the pubkey extracted, they pack into one slot, and the cost of reading or writing validator metadata drops to a single SLOAD or SSTORE. As-is, every committee enumeration, validator lookup during system calls, and reward-accounting path pays the multi-slot tax.

The pubkey also has a different access pattern from the rest of the struct. Most consensus paths read epoch boundaries, status, and stake version. Only proof-of-possession verification at staking time and committee-key fetches for off-chain consumers actually need the key bytes. Storing them together trades cheap reads of the hot fields against rare reads of the cold one.

## Solution

Move the BLS public key out of `ValidatorInfo` into a dedicated `mapping(address => bytes) blsPubkeys` keyed by validator address. With the dynamic field gone, the remaining members pack into one storage slot and validator-metadata reads collapse to a single SLOAD.

The registry should expose two view methods so callers that previously read the key through the struct retain a clean access path: `getBlsPubkey(address)` for a single validator, and `getCommitteeBlsPubkeys(uint32 epoch)` for the keys of an epoch's committee. The constructor should take the pubkey array as a separate argument parallel to the `ValidatorInfo[]` initializer, and the genesis arity check should cover the new array alongside the existing proofs-of-possession check.

Proof-of-possession verification, the staking flow, and the `blsPubkeyHashToValidator` reverse lookup continue to work as before. Only the storage layout and the function signatures around pubkey access change.

grantkee:
- `ValidatorInfo` no longer carries `bytes blsPubkey`. The key moves to `mapping(address => bytes) blsPubkeys`, so the remaining fixed-width members pack into a single storage slot
- Adds `getBlsPubkey(address)` and `getCommitteeBlsPubkeys(uint32)` to replace the field-access path
- Constructor takes a parallel `bytes[] blsPubkeys_` argument
    - the genesis arity check covers it alongside `proofsOfPossession`

closes #96


src/consensus/ConsensusRegistry.sol
```diff
@@ -203,6 +204,22 @@ contract ConsensusRegistry is StakeManager, Pausable, Ownable, ReentrancyGuard,
         return committeeValidators;
     }
 
+    /// @inheritdoc IConsensusRegistry
+    function getBlsPubkey(address validatorAddress) public view returns (bytes memory) {
+        return blsPubkeys[validatorAddress];
+    }
```
Huwonk, line 219: very minor nit, but might be nice to A) block zero address and B) revert if it's not a valid validatorAddress.  Non-blocking, but worth a look.

Huwonk (dismissed):
I left one comment related to `getBlsPubKey`  Definitely not a blocker.  Everything else looks great.

Huwonk (approved):
LGTM, good updates

src/consensus/ConsensusRegistry.sol
grantkee: Good idea - added

PR #98, opened 2026-04-30
Testnet Faucet Updates

Huwonk:
**Testnet Faucet Updates**
Updated deployments for testnet, using new salt.  

New faucet with enhanced access control functionality is in place and functional with a live test script to do 100% on-chain verification. `script\bash\test-faucet-drips.sh`  Updated storage, deprecated existing StablecoinManager for a clean install (not an in-place upgrade.)

**New Features**

- Drip limits per token
- Drip limit overrides per wallet per token
- Drip cooldown overrides per wallet
- Global drip cooldown stored in mutable variable
- List of all drippable tokens with metadata and drip limits for easy dApp integration
- Function to fetch user's next drip timestamp per token for easy dApp integration

Deprecated Step 6 in deploy-testnet-infra.sh.  The new deploy model now handles this natively via the StablecoinManager's constructor.

Implemented EXTCODESIZEGUARD fix as in grantkee's PR https://github.com/Telcoin-Association/tn-contracts/pull/95

100% line, branch and function coverage across StablecoinManager.sol and TNFaucet.sol


src/testnet/StablecoinManager.sol
```diff
@@ -64,64 +67,82 @@ contract StablecoinManager is StablecoinHandler, TNFaucet, UUPSUpgradeable {
...
         }
 
         _grantRole(DEFAULT_ADMIN_ROLE, initParams.admin_);
         _grantRole(MAINTAINER_ROLE, initParams.maintainer_);
     }
 
+    /// @notice Tombstone for the inherited 4-arg `UpdateXYZ`. Maintainers must use the 5-arg
+    ///         variant so every enabled token receives a baseline drip amount in the same call.
+    /// @dev Always reverts. The 5-arg overload below is the primary entry point.
+    function UpdateXYZ(address, bool, uint256, uint256) public pure override {
```
grantkee: nit: is this worth including? I think it makes sense if we were upgrading but I'm in favor of deploying fresh contracts for testnet

grantkee (dismissed):
LGTM! Thank you for being so thorough with documentation and tests. I had one nit, but not a show stopper. Great work

src/testnet/StablecoinManager.sol
Huwonk: yeah I'll remove this!

src/testnet/StablecoinManager.sol
```diff
@@ -64,64 +72,75 @@ contract StablecoinManager is StablecoinHandler, TNFaucet, UUPSUpgradeable {
...
+    ///      set, forwards `(maxLimit, minLimit)` to StablecoinHandler, and seeds the faucet's
+    ///      baseline drip amount via `_setMaxDripAmount` (which itself rejects 0).
+    /// @dev On `validity == false`: removes the token from the drippable set and forwards to
+    ///      StablecoinHandler. `baseDripAmount` is ignored on this path.
+    /// @param token Token address to enable or disable. Use `NATIVE_TOKEN_POINTER` for native.
+    /// @param validity True to enable, false to disable.
+    /// @param maxLimit StablecoinHandler max mint cap.
+    /// @param minLimit StablecoinHandler min burn cap.
+    /// @param baseDripAmount Baseline max drip amount for this token (only consumed on enable).
     function UpdateXYZ(
```
chasebrownn, line 117: So a weird issue I found is that if you disable a token drip, but then try to re-enable it with the same settings -> it will revert since `_baselineMaxDripAmount` is never cleared. Here's a test case to confirm:
```solidity
function testReEnableSameBaselineReverts() public {
        vm.startPrank(maintainer);

        // 1. Enable token1
        stablecoinManager.UpdateXYZ(token1, true, 1000, 1, 100);
        assertEq(stablecoinManager.getBaselineMaxDripAmount(token1), 100);

        // 2. Disable token1
        //    so _baselineMaxDripAmount[token1] stays at 100
        stablecoinManager.UpdateXYZ(token1, false, 1000, 1, 0);
        assertFalse(stablecoinManager.isEnabledXYZ(token1));

        // 3. Re-enable with the same settings -> reverts via SettingAlreadyConfigured
        vm.expectRevert(TNFaucet.SettingAlreadyConfigured.selector);
        stablecoinManager.UpdateXYZ(token1, true, 1000, 1, 100);

        vm.stopPrank();
    }

Ran 1 test for test/faucet/StablecoinManager.t.sol:StablecoinManagerTest
[PASS] testReEnableSameBaselineReverts() (gas: 194640)
```
An odd edge case and I'm sure if you disable a drip token you probably wont re-enable it. And if you do we can assign it slightly different settings as a workaround. Not sure if this is desired behavior though or maybe I'm misusing the implementation.

src/testnet/StablecoinManager.sol
```diff
@@ -141,103 +160,115 @@ contract StablecoinManager is StablecoinHandler, TNFaucet, UUPSUpgradeable {
...
-        if (block.timestamp < lastFulfilledDrip + 1 days) {
-            revert RequestIneligibleUntil(lastFulfilledDrip + 1 days);
-        }
+        // cooldown check, applies per-recipient override if set
+        uint256 nextEligibleAt = getNextEligibleDripTimestamp(recipient, token);
+        if (block.timestamp < nextEligibleAt) revert RequestIneligibleUntil(nextEligibleAt);
+
+        // amount check, applies per-recipient override if set
+        uint256 maxAmount = getMaxDripAmount(recipient, token);
+        if (amount > maxAmount) revert RequestedAmountTooHigh(amount, maxAmount);
```
chasebrownn, line 228: Not sure if this is the intended design model, if it is then just ignore me. But this current drip model is "one claim per window" no matter the amount. Which means if I claim 1 wei or maxAmount, I'll be locked out of claims for the next 24 hours (or whatever the cooldown period is). Is this intended? Feels weird not being able to claim the entire amount allowed during that cooldown window. Technically this does open a griefing vector where if an attacker wanted to block Noah's wallet of claiming everything allotted to him he could just claim on his behalf everytime the window open 1 wei, blocking him from claiming the max amount he can claim. Now this is trivial because Noah can also just create more wallets, but just food for thought.

chasebrownn (commented):
No deal breakers, but a couple notes worth a quick look.

src/testnet/StablecoinManager.sol
Huwonk: Yeah so I asked about dripTo in general because of the same issue, so this is a solid point.  Maybe we can just make it so the requested amount needs to at least be 10% of the minimum or something.

src/testnet/StablecoinManager.sol
Huwonk: Good catch.  Ideally it'd zero out if we set validity to false.  I'd say it's a bug, but a very minor one.  I can create some follow-up issues.  It's not deployed during Genesis and it is behind a proxy so we can upgrade it.

chasebrownn (approved):
Good stuff @Huwonk . Looks good.

PR #99, opened 2026-05-01
Bug/fix ci execution - Important

Huwonk:
**Fixed bug with CI having zero execution success in Github.**

- Removed `forge fmt` from Github Actions CI - We can decide if we want to add this back in later, but it's out of scope for this PR.
- Removed the --sizes parameter from forge build - ConsensusRegistry.sol fails because it is a precompile.  
- Reduced fuzz runs for base CI to 256 from 10000
- Added new Foundry profile for attestation for the 10000 fuzz run which can only be executed locally.  GH Actions time out on this. 

**What does work**

- `forge build`
- `forge test --vvv`

Must merge before rebasing new PRs to set a new baseline for CI across the repo.


chasebrownn (approved):
LGTM

PR #101, opened 2026-05-04
add 0xfe to precompile in genesis

### LINKED ISSUE #102: Genesis state for TEL precompile breaks Solidity typed-interface calls
## Problem

The TEL native ERC20 precompile lives at `0x00000000000000000000000000000000000007e1` and is dispatched at runtime by revm on the Rust client side — no bytecode is ever deployed for it on-chain. As a result, the precompile's account exists in genesis state with `code = 0x` and `EXTCODESIZE` returns `0`.

Solidity's compiler emits an implicit `EXTCODESIZE` check in front of every typed external call (`IFoo(addr).bar()`). When that check sees an empty account, the call reverts before any value is forwarded — and before revm has a chance to route the call to the native precompile handler. Any contract that tries to interact with TEL through a typed interface therefore fails, even though the precompile is fully functional when reached.

This has already produced one workaround in production code: `StablecoinManager::_drip` calls the precompile through a manually constructed `address.call(abi.encodeWithSelector(...))` to bypass the `EXTCODESIZE` guard. Every future contract that needs to call TEL — faucets, governance flows, integrations from third parties — would have to repeat that low-level pattern, which is fragile, easy to get wrong, and hostile to auditors who reasonably expect typed interfaces to work against an account documented as a standard ERC20.

The root cause is at the genesis layer, not in any individual contract. As long as the precompile's genesis account has empty code, every Solidity caller has to know an implementation detail of revm's dispatcher.

## Solution

The genesis configuration should seed the TEL precompile account with a single byte of placeholder code so that `EXTCODESIZE` returns non-zero and Solidity's typed-interface guard passes. The byte chosen should be `0xfe` — the EVM `INVALID` opcode — so that if any execution path were ever to actually run the bytecode (which should not happen, since revm intercepts the address before EVM dispatch), it fails closed rather than executing accidental behavior.

Two places need the change:

- `deployments/genesis/precompile-config.yaml` — the canonical genesis precompile config consumed by node operators — should include an entry for `0x7e1` with `code: 0xfe` alongside the existing system precompiles (Multicall3, CREATE2 factory, EIP-2935, EIP-4788, the governance Safe).
- `script/GenerateGenesisPrecompileConfig.s.sol` — the Foundry script that regenerates the yaml — should emit the same entry, so the file and the generator stay in sync and future regenerations don't drop the precompile back to empty code.

This change is purely additive at the genesis layer. The TEL precompile's runtime semantics are unchanged: revm dispatches the address before EVM execution, so the `0xfe` byte is never run. Existing workarounds (such as `StablecoinManager`'s low-level `.call`) continue to function unchanged but become unnecessary for new contracts. Auditors and integrators see a precompile address that behaves like any other contract from the caller's perspective, removing a class of footguns from the integration surface.

grantkee:
The TEL precompile at 0x7e1 is a native Rust precompile (not EVM bytecode). When Solidity contracts make high-level calls to it (e.g., ITELMint(0x7e1).transfer(...)), the compiler inserts an EXTCODESIZE check that reverts if the target has no code. Without a genesis account entry, this check fails.

The fix: add a genesis account at 0x7e1 with code: 0xfe (the INVALID opcode, 1 byte). This makes EXTCODESIZE return 1, satisfying the check, while the actual call is intercepted by the native precompile handler before the EVM ever tries to execute the bytecode.

closes #102

### DIFF
--- a/deployments/genesis/precompile-config.yaml
+++ b/deployments/genesis/precompile-config.yaml
@@ -56,4 +56,4 @@
 "0x00000000000000000000000000000000000007e1": # TEL precompile
   nonce: 0
   balance: 0
-  code: 0xfe
+  code: "0xfe"
--- a/deployments/genesis/precompile-config.yaml
+++ b/deployments/genesis/precompile-config.yaml
@@ -53,3 +53,7 @@
 "0x3fAB184622Dc19b6109349B94811493BF2a45362": # use nonce 0 for creating 0x4e59b44847b379578588920cA78FbF26c0B4956C
   nonce: 1
   balance: 0
+"0x00000000000000000000000000000000000007e1": # TEL precompile
+  nonce: 0
+  balance: 0
+  code: 0xfe
--- a/script/GenerateGenesisPrecompileConfig.s.sol
+++ b/script/GenerateGenesisPrecompileConfig.s.sol
@@ -90,6 +90,9 @@ contract GenerateGenesisPrecompileConfig is GenesisPrecompiler, Script {
         // Arachnid deterministic deployment proxy (CREATE2)
         instantiateArachnidFactory();
 
+        // TEL precompile (native Rust handler at 0x7e1, needs code for EXTCODESIZE check)
+        instantiateTelPrecompile();
+
         vm.stopBroadcast();
     }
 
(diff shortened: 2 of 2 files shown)


Huwonk (dismissed):
Yeah this is exactly what I had thought.  Looks good.

Huwonk (approved):

PR #105, opened 2026-05-07
Worker configs strategy

### LINKED ISSUE #104: Refactor WorkerConfigs for Protocol Use
## Problem

`WorkerConfigs` is the on-chain source of truth for the per-worker fee policy that the protocol layer reads at every epoch boundary. Each entry holds a `(strategy, value)` pair, where `strategy` selects how the protocol interprets `value` (currently `0 = EIP-1559`, `1 = Static`, mirroring `WorkerFeeConfig` in `tn-types::gas_accumulator`).

Today the contract validates these fields the wrong way around:

- **`strategy` is not validated at all.** Owners can write any `uint8` (0–255) through the constructor or the setters. Unknown strategy ids only fail later, inside the Rust protocol code, after the configured value has already been accepted on-chain. A misconfiguration that should be caught at the contract boundary instead surfaces as a runtime decode error against committed state.
- **`value` is required to be `≥ MIN_GAS = 7 wei`**, and `setNumWorkers` uses `value < MIN_GAS` as a proxy for "this worker has no config yet". That conflates two unrelated concerns. A `Static{fee: 0}` config — a legitimate policy where the worker charges no base fee — is rejected purely because the contract uses the value field as a coverage flag. Any future strategy that legitimately wants a small or zero value runs into the same wall.

A second, lesser issue: reading the full fee policy requires `numWorkers + 1` separate EVM transitions (one `numWorkers()` call plus one `getWorkerConfig` per worker). The protocol layer always wants the whole set together, so the loop on the Rust side is pure overhead.

## Solution

`WorkerConfigs` should validate the field that has a defined domain (`strategy`) and stop mis-using the field that doesn't (`value`).

Introduce a `MAX_STRATEGY` constant on the contract that mirrors the highest variant of `WorkerFeeConfig` on the Rust side, and a corresponding `InvalidStrategy(uint8)` error. The constructor, single setter, and batch setter should all reject `strategy > MAX_STRATEGY`, so an unknown strategy id can never reach the protocol layer through the configured path. When new strategies are added, `MAX_STRATEGY` moves in lockstep with the Rust enum and the contract upgrade.

Coverage tracking should move off the value field and onto an explicit "has this worker been configured?" flag. An internal `_workerConfigSet` mapping, written on every successful set, lets `setNumWorkers` reject worker ids that were never configured without making any claim about what their value should be. With coverage decoupled from value, `MIN_GAS` no longer serves a purpose; it should be retained as a `uint64` constant set to `0` for ABI continuity (external tooling reads it), but the `ValueBelowMinGas` error and every site that threw it should be removed.

Finally, `WorkerConfigs` should expose a single `getAllWorkerConfigs()` view that returns `(count, strategies[], values[])` so the protocol layer can fetch the full per-epoch fee policy in one EVM call.

The tradeoff worth flagging: keeping `MIN_GAS` in the ABI as a `0`-valued constant is mildly misleading — it implies a floor that the contract no longer enforces. The natspec is updated to make the new semantics explicit, and the alternative (a breaking ABI change) would invalidate the embedded bytecode that the Rust genesis path loads from `artifacts/WorkerConfigs.json`.

grantkee:
- Add `MAX_STRATEGY` (currently `1`, mirroring `WorkerFeeConfig` on the Rust side) and `InvalidStrategy(uint8)`. Constructor and both setters reject any `strategy > MAX_STRATEGY` so unknown ids cannot reach the protocol layer.
- Replace the `value < MIN_GAS` coverage proxy with an explicit `_workerConfigSet` mapping. `Static{fee: 0}` and similar legitimate zero-valued configs are now legal at construction, single-set, and batch-set.
- `MIN_GAS` is kept as a `uint64` constant pinned to `0` for ABI continuity; `ValueBelowMinGas` and every site that threw it are removed. `MissingWorkerConfig` natspec is reworded around "never explicitly set".
- Add `getAllWorkerConfigs() -> (count, strategies[], values[])` so the protocol can read the full per-epoch fee policy in one EVM call instead of `numWorkers + 1`.
- Refresh `artifacts/WorkerConfigs.json` so the bytecode the Rust genesis path embeds exposes the new selectors.
- Tests: drop the 7-wei floor cases, add positive coverage for zero values, tighten fuzz strategies via `vm.assume(strategy <= MAX_STRATEGY)`, and add positive/negative cases for the new bounds check on each entry point.

closes #104


Huwonk (approved):
Looking good after some back and forth

PR #106, opened 2026-05-07
fix precompile script

grantkee:
- use string literals for precompile code to prevent yaml from deserializing `0xfe` as an int


Huwonk (approved):
weird I thought we had fixed this.

PR #108, opened 2026-06-01
Pin BlsG1 library to canonical address 0xb151 in tests

Huwonk:
Link the BlsG1 library to a fixed vanity address (0xb151) instead of an ephemeral per-run one, so the test harness exercises the same layout the network uses at genesis. This is groundwork for eventually reimplementing BlsG1 as a native Rust precompile at the same address, which will be address-stable for all consumers.

- foundry.toml: pin BlsG1 via `libraries`. Forge no longer auto-deploys the library when an address is pinned, so tests place its bytecode at the address themselves. Add a `profile.artifacts` that clears the pin so the vendored artifacts/ keep the link placeholder the protocol fills at genesis (regenerate with `FOUNDRY_PROFILE=artifacts forge build`).
- BlsG1.sol: add `BLS_G1_ADDRESS` constant as the single source of truth (unused constant, so no bytecode change).
- Deployments.sol / deployments.json: record the address as `MagicAddresses.BLS_G1_LIBRARY` (lexicographically first so the struct still decodes).
- test/EIP2537/BlsG1Deployed.sol: shared base that etches the library at the pinned address; EIP2537 suites inherit it.
- ConsensusRegistryTestUtils: the util builds a genesis registry in the test contract's own constructor (before any test setup can etch). Make ConsensusRegistry._verifyProofOfPossession virtual (deployed bytecode is byte-identical) and override it to skip BLS for the unused inherited self-instance, pass BLS-free dummy genesis args, and defer the real PoPs and validator5 key to the constructor body where the library is etched.

Full suite green (256 passed, 2 fork-skips) under the pinned profile.


chasebrownn (approved):
Looks good

grantkee (approved):

PR #109, opened 2026-06-01
Consolidate ConsensusRegistry proof-of-possession into a single BlsG1 call

Huwonk:
## Summary

Second PR in the BlsG1-precompile prep series, stacked on #108. This cleans up the
handoff from `ConsensusRegistry` to the `BlsG1` library so a validator proof of
possession is verified in a single library call, ahead of moving `BlsG1` to a native
Rust precompile.

## What changed

- `ConsensusRegistry._verifyProofOfPossession` no longer orchestrates encoding, message
  assembly, and the pairing check across multiple library calls. It hands the raw inputs
  to `BlsG1.verifyProofOfPossession(signature, pubkey, address)` once and gets a bool back.
  Runtime delegatecalls from ConsensusRegistry into BlsG1 drop from 5 sites to 2.
- The proof-of-possession scheme now lives entirely in `BlsG1` (prefixes, message
  assembly, EIP2537 encoding, hash-to-curve, pairing check) as the single source of truth
  and the eventual precompile's internal logic.
- Parameter order mirrors the protocol's Rust
  `verify_proof_of_possession_bls(proof, public_key, address)` so the call site is stable
  when the implementation moves to the native precompile. Both crypto params are `bytes`,
  so the selector is order-agnostic and the order must match semantically.
- `encodeG2/encodeG1/verifyProofOfPossessionG1` change from `external` to `public` so the
  new function composes them as in-library jumps (one delegatecall from ConsensusRegistry).
- Vendored `artifacts/` for `ConsensusRegistry` and `BlsG1` regenerated with the
  placeholder-preserving `artifacts` profile, so telcoin-network consumes the new bytecode
  with the `__$..$__` link placeholder intact (not a baked address).

## What did not change

- ConsensusRegistry's external ABI and the network's inbound messages are untouched: the
  `stake` and genesis inputs keep their shapes, and `proofOfPossessionMessage` stays as a
  thin forwarder to `BlsG1`.

## Testing

- Full suite green: 256 passed, 0 failed, 2 fork-skips, under the pinned profile.
- `testRevert_stake_invalidPoint` now asserts the more descriptive `InvalidBLSPubkey` and
  its signature input is corrected to 96 bytes.

## Notes

- Stacked on #108; that one should land first. GitHub will retarget this PR to `master`
  once #108 merges.


src/consensus/BlsG1.sol
```diff
@@ -238,6 +244,52 @@ library BlsG1 {
...
+    /// @param uncompressedPubkey 192-byte uncompressed G2 public key
+    /// @param validatorAddress The validator's execution address bound into the PoP message
+    /// @return _ True if the proof of possession is valid
+    /// @dev Parameter order mirrors the protocol's Rust `verify_proof_of_possession_bls(proof, public_key,
+    /// address)` so this call site is stable when the implementation moves to the native precompile.
+    function verifyProofOfPossession(
+        bytes memory uncompressedSignature,
+        bytes memory uncompressedPubkey,
+        address validatorAddress
+    )
```
chasebrownn, line 281: nit: Might be worth swapping these 2 arguments. The original `ProofOfPossession` struct, `verifyProofOfPossessionG1`, and `proofOfPossessionMessage` all take `Pubkey` first. It's trivial, but for DX (developer experience - new term) might be worth aligning with 
```solidity
function verifyProofOfPossession(
        bytes memory uncompressedPubkey,
        bytes memory uncompressedSignature,
        address validatorAddress
    )
```
to avoid having to reverse arguments for this 1 call.

chasebrownn (dismissed):
Might be worth just swapping the arguments in `verifyProofOfPossession` for consistency. Other than that - looks beautiful.

src/consensus/BlsG1.sol
Huwonk: Opening a convo with @grantkee tomorrow RE: this.

grantkee (approved):
LGTM! Good work @Huwonk

PR #112, opened 2026-06-02
ConsensusRegistry BLS proof-of-possession fixes + audit cleanups (#110, #111)

### LINKED ISSUE #110: In ConsensusRegistry, fix ordering of parameters
Currently the downstream Bls library will expect signature, pubkey, address.  We currently have pubkey, signature, address in several areas.  Clean-up item before mainnet.

### LINKED ISSUE #111: In ConsensusRegistry, add check to confirm blsPubKey and uncompressPubKey are aligned

Huwonk:
Pre-mainnet ConsensusRegistry BLS proof-of-possession hardening (#110, #111), plus a review
follow-up (EXP-001) and two low-severity audit cleanups in the same file.

Closes #110
Closes #111

## #110 - canonical `(signature, pubkey, address)` ordering

Aligns the BLS verify surface to the signature-first order used by the protocol's Rust
`verify_proof_of_possession_bls(proof, public_key, address)`, so the call site is stable when the
implementation moves to the native precompile.

- `BlsG1.verifyProofOfPossessionG1` reordered to `(signature, blsPubkey, message, dst)`; in-library
  call and all test call sites updated. The pairing-input concatenation inside the function is
  unchanged (fixed by the pairing equation, not a parameter).
- `BlsG1.verifyProofOfPossession` was already signature-first.
- The `ProofOfPossession` struct is intentionally left pubkey-first: Solidity accesses its fields by
  name, so struct order does not affect call ordering, and leaving it avoids changing the
  validator-facing stake/delegateStake calldata encoding or the Rust genesis encoder.

## #111 - bind the registered key to the proven key (security)

`_verifyProofOfPossession` proves possession of `g1Pop.uncompressedPubkey`, but the consensus key
stored and aggregated by the protocol is the separate compressed `blsPubkey`. Nothing bound the two,
so a validator could register a consensus key for which no proof of possession exists. Committee
certificates are verified with same-message BLS aggregate verification (FastAggregateVerify), which
relies on registration-time proof of possession to prevent rogue-key attacks - so the gap let a
single registered rogue key forge an aggregate that appears signed by a quorum.

Fix: an on-chain x-coordinate match between the compressed and uncompressed keys, in `ConsensusRegistry`
(the consumer) so it does not depend on the eventual precompile honoring the binding. blst serializes
G2 as `x.c1 || x.c0 || y.c1 || y.c0` (uncompressed) and `x.c1 || x.c0` (compressed, 3 flag bits in byte
0); we clear the flag bits and compare the 96-byte x. A rogue cancellation key has a different x than
any key the caller can prove, so an x match is sufficient. Mismatches revert `BLSPubkeyMismatch`.
Validated against the canonical blst-compressed BLS12-381 G2 generator (a real vector, independent of
the harness's derived keys), plus a negative test.

## EXP-001 follow-up - deduplicate by x-coordinate (security)

Review of #111 surfaced a related gap. The x-match binds the stored key to the proven key, but
deduplication still hashed the full compressed bytes - so a key and its y-sign negation (same x, byte
0's sign flag differs) registered as two distinct validators, violating UNIQ-BLS-KEYPAIR. Two same-x
validators sampled into one committee sum to the identity point, enabling a FastAggregateVerify
denial-of-service, and inflate Sybil committee weight.

Fix: a new `_blsKeyId` hashes the x-coordinate (compressed key with byte 0's flag bits cleared);
`_spendBLSPubkey`, `isValidator`, and `delegationDigest` all key on it (the last must move in lockstep
or delegated staking fails signature verification). Two distinct valid G2 points share an x only if
they are negations, so this collapses the +/- variants into one ledger entry (the second hits
`DuplicateBLSPubkey`) while never colliding honest distinct validators. Regression test included.

The two checks compose: x-match (anti rogue-key, different x) + x-dedup (anti negation, same x). Gas:
the x-keyed id adds ~315 gas over the prior keccak on a once-per-validator path (~0.06% of `stake()`).

## Audit cleanups (low severity)

- `_enforceSorting` iterated `i < length - 1`, underflowing to `type(uint256).max` on an empty
  committee (array out-of-bounds panic). Now iterates from index 1; an empty committee is rejected
  cleanly by `concludeEpoch`'s length check (`InvalidCommitteeSize`). `onlySystemCall`, so it was not
  attacker-reachable, but it removes a latent footgun. Regression test added.
- `IConsensusRegistry.proofOfPossessionMessage` declared `view` while the implementation is `pure`;
  synced the interface to `pure`.

## Audit findings disposition

The review covered 7 findings:

| # | Finding | Status |
|---|---------|--------|
| 1 | PoP key decoupling (High) | Fixed (#111 x-match + EXP-001 x-dedup) |
| 2 | `_enforceSorting` empty-input underflow (Low) | Fixed |
| 3 | Committee sort order not preserved after ejection (Low) | Open - needs protocol confirmation (does consensus rely on sorted committees after `_eject`'s swap-and-pop?) |
| 4 | `_getValidators` O(n) scans per epoch (Info) | Acknowledged future work (SSTORE2); ~1000-validator ceiling is a hard bound |
| 5 | `proofOfPossessionMessage` interface `view` vs `pure` (Info) | Fixed |
| 6 | Owner centralization powers (Info) | By design (governance Safe) |
| 7 | CR / Issuance solvency coupling (Info) | Out of scope (`Issuance.sol` review) |

Only Finding 3 remains actionable and it is a question rather than a known defect: if the protocol
relies on stored committee ordering after ejection, `_eject` needs an order-preserving removal;
otherwise it is fine as-is.

## Notes

- Vendored `artifacts/{ConsensusRegistry,BlsG1}.json` regenerated with the placeholder-preserving
  `artifacts` profile, so the `__$..$__` link placeholder is kept (filled at genesis) rather than a
  baked address. ABI changes are additive (`BLSPubkeyMismatch`); `verifyProofOfPossessionG1`'s selector
  is unchanged (parameter types identical, only order moved).
- Full suite green: 260 passed, 2 fork-skips.


chasebrownn (dismissed):
luks gud

utkutugrul:
## Independent security review — ConsensusRegistry BLS PoP hardening (#110/#111/EXP-001)

I ran an independent review of this PR (commit `9b56260`) and wanted to share the results. **TL;DR: this is a sound hardening — no Critical/High/Medium issues found, and the three changes are correct.** A few Low/Informational suggestions and two off-chain confirmation items are below.

> Transparency: this is an AI-assisted review (static analysis + property fuzzing + symbolic execution + manual cryptographic review). Sharing for your consideration; please treat the suggestions as optional and verify independently.

### What was verified correct
- **#111 x-coordinate binding (`_blsKeysAligned`)** — correctly closes the rogue/cancellation-key gap. Two valid G2 points share an x only as `±P` (both prover-controlled), so binding the stored compressed key to the PoP-proven key by x is sufficient. The flag-mask `0x1f…` clears exactly the 3 blst flag bits, and since the field prime's top byte is `0x1a < 0x20` it never touches real x data. Reads are length-safe (`blsPubkey == 96` at L614; `uncompressedPubkey == 192` enforced by the pairing path at L618, both before the assembly at L626).
- **EXP-001 x-dedup (`_blsKeyId`)** — collapses a key and its y-sign negation to one id, blocking negation double-registration (the `P + (−P) = identity` committee-cancellation vector). Keying is consistent across **all** consumers including the genesis path (`isValidator`, `delegationDigest`, `_spendBLSPubkey`, genesis `_verifyProofOfPossession`) — no stale `keccak256(blsPubkey)` remains.
- **`_enforceSorting` underflow fix** — correct; empty committee is cleanly rejected by `_checkCommitteeSize` (and the path is `onlySystemCall`, so the old underflow was never attacker-reachable).
- **Parameter reorder (#110)** — pure relabel; pairing-input order and selector unchanged.

Evidence: 116/116 consensus + 58/58 EIP-2537 tests pass; a differential property-fuzz of the real `_blsKeyId`/`_blsKeysAligned` (Foundry 250k + Echidna 500k + Medusa 570k iterations, all green); and **halmos symbolically proved** 4 properties (id↔reference equality, flag-bit invariance, negation collapse, alignment correctness).

### Low / Informational suggestions (optional)
- **[Low] Stored compressed key flag bits are not validated on-chain.** `_blsKeysAligned`/`_blsKeyId` mask the compression/infinity/sign bits and bind only the x-coordinate. A whitelisted validator could store a `blsPubkey` whose x matches its proven key but whose flag bits are malformed (e.g. infinity bit set → off-chain blst decodes to identity; compression bit cleared → decode error). On-chain this is **self-limited** (x-dedup + NFT gate + PoP mean it can't grief another validator), but the worst case depends on how the off-chain Rust/blst layer handles a malformed committee member (excluded vs. whole-epoch abort). Suggestion: either confirm the off-chain decoder rejects malformed/identity keys, or add a cheap on-chain guard asserting byte-0's compression bit is set and infinity bit clear.
- **[Info] `delegationDigest` lacks the 96-byte length guard** that `isValidator` has (it calls `_blsKeyId` which `mload`s a fixed 96 bytes). It's inert in practice — `delegateStake` re-derives the id from a length-checked `bytes calldata`, so a malformed-length digest just fails signature verification — but a one-line `if (blsPubkey.length != 96) revert BlsG1.InvalidBLSPubkey();` would restore parity for off-chain integrators.
- **[Info] `_blsKeysAligned` masks the uncompressed operand too.** This is correct only because the pairing check (which guarantees the uncompressed key's top bits are zero) runs first. Consider not masking the uncompressed operand (or asserting its top 3 bits are zero) so the check stays correct independent of call ordering.
- **[Info] Compiler pin.** solc 0.8.28–0.8.33 carry a transient-storage `tstore` bug under `--via-ir`; you're on 0.8.26 with Solady's non-transient `ReentrancyGuard` so you're **not affected**, but worth pinning a known-good version and avoiding that range.

### Two off-chain confirmation items (not contract bugs)
- **PoP message parity:** the on-chain `proofOfPossessionMessage` (address-bound, uncompressed pubkey, prefix `0x000000d501`) should be byte-identical to the genesis Rust verifier / planned precompile. This isn't changed by the PR; just worth confirming the two sides agree on the preimage.
- **Committee ordering after `_eject`:** this matches your own open "Finding 3" — `_eject`'s swap-and-pop unsorts the stored committees. No on-chain consumer depends on stored ordering (all readers are linear scans; sorting is only enforced on the system-supplied `futureCommittee` for duplicate-prevention), so it's safe on-chain; the only question is whether off-chain consensus assumes sorted stored committees.

Nice work on the hardening — the negation-dedup composition (x-match for different-x rogue keys + x-dedup for same-x negations) is a clean way to cover both cases, and the real-blst-generator test vector + the new regression tests are a good touch.

Huwonk:
Thanks @utkutugrul, this is a thorough and useful review. We pushed `28b048a` on this branch taking two of the suggestions, and have reasoning for deferring the rest.

### Taken
- **[Low] stored compressed-key flag bits.** Added a guard in `_verifyProofOfPossession` requiring byte 0's compression flag set and infinity flag clear:
  ```solidity
  uint8 flags = uint8(blsPubkey[0]);
  if (flags & 0x80 == 0 || flags & 0x40 != 0) revert BlsG1.InvalidBLSPubkey();
  ```
  We went on-chain rather than relying on the off-chain decoder because the worst case (infinity flag set -> identity committee member, or compression flag cleared -> decode failure) is reachable by a whitelisted validator and could abort epoch processing off-chain, and the Rust/blst decoder behavior for malformed members isn't finalized yet (the precompile/library is still ahead of us). We confirmed against telcoin-network `bls_signature.rs` that blst's compressed serialization always sets compression and clears infinity for real keys, so genesis and stake are unaffected. We left the sign flag (`0x20`) free intentionally: a wrong sign only breaks that validator's own signatures, which BFT tolerates, and binding it on-chain would mean recomputing the canonical y-sign rule, which we'd rather leave to the off-chain/precompile layer.
- **[Info] `delegationDigest` length guard.** Added the `blsPubkey.length != 96` check for parity with `isValidator`/`_verifyProofOfPossession`.

Both are covered by a new regression test (`testRevert_stake_malformedCompressedFlags`) exercising the infinity-set and compression-cleared cases.

### Deferred (with reasoning)
- **[Info] `_blsKeysAligned` masking the uncompressed operand.** Leaving as-is. The function is only ever reached after the pairing check has already established the uncompressed key is a valid Fp element (top bits zero), so the mask on that operand is a no-op for every real input. Since you symbolically verified the current assembly, we'd rather not alter it for a hypothetical future re-ordering and invalidate that proof. If we ever add a caller that runs alignment before the pairing, we'll revisit with an explicit top-bits-zero assertion.
- **[Info] compiler pin.** Already covered: `foundry.toml` pins `solc = "0.8.26"` with `auto_detect_solc = false`, no `via_ir`, and Solady's non-transient `ReentrancyGuard`, so the 0.8.28-0.8.33 `tstore` issue is out of scope.
- **Off-chain items (PoP message parity, committee ordering after `_eject`).** Both are confirmation items rather than contract changes. The `_eject` ordering matches our existing "Finding 3" (no on-chain consumer depends on stored committee order; sorting is only enforced on the system-supplied `futureCommittee`), so it's safe on-chain and the only open question is whether off-chain consensus assumes sorted stored committees. We're tracking both alongside the precompile migration, where the PoP preimage and decoder behavior get pinned against the Rust side.

Appreciate the halmos run and the real-blst-generator framing - that gave us good confidence on the x-binding composition.

grantkee (approved):
LGTM! Great work

PR #113, opened 2026-06-03
O(1) committee-eligible validator counter

Huwonk:
**WIP / draft for team review.** Stacked on `fix/consensus-registry-bls-pop` (the BLS PoP + x-coordinate dedup work, #110/#111). Review just the top commit; this only touches the contract-side count accounting.

## What

The four hot count sites all ran `_getValidators(Active).length`:

- `concludeEpoch` (every epoch)
- `setNextCommitteeSize`
- `beginExit`
- `_consensusBurn`

`_getValidators(Active)` allocates a `totalSupply()`-sized array and cold-SLOADs every validator's slots, purely to learn the size of the unretired `{ PendingActivation, Active, PendingExit }` set. We replace those reads with `eligibleValidatorCount`, a single `uint256` maintained incrementally.

## How it stays correct

- A new private `_setStatus(validator, newStatus)` helper does every `currentStatus` write and adjusts the counter by comparing committee-eligibility before/after. `_beginActivation`, `_activate`, `_beginExit`, `_exit` all route through it.
- `_retire` decrements if the validator was still eligible (defensive against any retire-without-exit path; idempotent because it checks `!isRetired && eligible(currentStatus)`).
- The constructor seeds `eligibleValidatorCount = initialValidators_.length` (every genesis validator is Active).

The committee-eligible set is exactly `{ PendingActivation, Active, PendingExit }` among unretired validators, which is exactly what `getValidators(Active)` returns, so the counter tracks that length without scanning.

## Why no telcoin-network change is needed

`getValidators(Active)` and its **array order** are left completely untouched. `block.rs::shuffle_new_committee` reads `getValidators(Active)` and Fisher-Yates shuffles it, so the array order is consensus-critical; we did not move that. Only the order-irrelevant **counts** move to the counter.

## Gas

Synthetic genesis sets, `getEligibleValidatorCount()` vs `getValidators(Active)`:

| N validators | O(1) counter | O(n) scan |
|---|---|---|
| 8 | 729 | 29,227 |
| 32 | 729 | 112,157 |
| 64 | 729 | 224,424 |
| 128 | 729 | 454,767 |

Flat 729 gas vs ~3,550 gas per validator. At 1,000 validators that is ~3.5M gas per scan, paid every epoch in `concludeEpoch`.

## Tests

- `getEligibleValidatorCount()` getter
- `_assertEligibleInvariant()` (counter == `getValidators(Active).length`) hooked into the heaviest fuzz tests (`testFuzz_mintStakeBurn`, `testFuzz_concludeEpoch_success`)
- `test_eligibleValidatorCount_burnDecrements` lifecycle test
- `ConsensusRegistryGasBench` benchmark (heavyweight; happy to gate or drop before merge)
- Full consensus suite passes (74 tests)

## Open questions for review

- Keep the gas benchmark in CI, or move it behind a profile / drop it?
- Anything in the protocol that reads the eligible **count** from a path we have not considered?


grantkee (approved):
LGTM! Great idea

PR #114, opened 2026-06-05
Index validators by per-status EnumerableSet

Huwonk:
**Ready for review.** Targets master directly (no longer stacked); synced with master (solc 0.8.35, regenerated artifact). Supersedes the O(1)-counter work that already merged as #113.

## What

Index validators by one OpenZeppelin `EnumerableSet.AddressSet` per `ValidatorStatus`, keyed by the enum, replacing the O(totalSupply) `tokenByIndex` scan in `_getValidators` with a single set read.

- `_setStatus` is the one place that mutates membership (remove old status set, add new) **and** maintains the cached `eligibleValidatorCount`, keeping both in lockstep with `validators[].currentStatus`. `_recordStaked`, `_retire`, and the genesis loop route through it (or seed explicitly).
- The eligible count stays a cached single SLOAD; the sets back the address queries. The membership invariant verifies the cache against the sets.
- `getValidators(status)` now returns `address[]` (the exact status set; reverts on `Undefined`/`Any`); the `Active` query no longer folds in `PendingActivation`/`PendingExit` (the protocol unions the three eligible queries off-chain, the on-chain count is that union's size). New `getValidatorsInfo(status)` returns `ValidatorInfo[]`.

Protocol-side changes are specified in `src/consensus/status-sets-protocol-migration.md`.

## Also includes the #115 safe wins (folded from the former #116)

- Extracted the triplicated stake-originator access-control preamble into `_checkStakeOriginator(address) returns (recipient)` (used by `claimStakeRewards`, `upgradeValidatorStakeVersion`, `unstake`).
- Removed the now-dead `StakeManager._getAddress`.
- `beginExit` reads `validator.activationEpoch` via the existing storage ref.

The deferred #115 items (decomposing `upgradeValidatorStakeVersion`, the dormant reward scaffolding, optional micro-gas) remain tracked in #115.

## Gas (synthetic genesis, N = 8 / 32 / 64 / 128)

| op | before (scan) | after (set) |
|---|---|---|
| `getValidators(Active)` (protocol path) | ValidatorInfo[] 29,227 / 112,157 / 224,424 / **454,767** | address[] **5,478 / 15,206 / 28,192 / 54,212** |
| `getEligibleValidatorCount` | 729 | 766 (cached SLOAD, kept) |

## Tests

Full suite: **268 passing** (post master sync). Membership invariant (`_assertSetInvariant`) hooked into the fuzz tests. Differential audit found no Critical/High/Medium.

Refs #115.


chasebrownn (approved):
LGTM

PR #118, opened 2026-06-08
Bound exit-queue gas with a transient committee-membership set

Huwonk:
**Ready for review.** Based on `enhancement/consensus-registry-status-sets` (#114) so it carries the set refactor and merges cleanly; review #114 first - this will be retargeted to master once #114 lands. Synced with the master-synced base (solc 0.8.35).

## What

`_updateValidatorQueue` tested each pending-exit validator against three committees with a linear scan - `O(P x 3C)`, where P = exit-queue depth and C = committee size (up to 100). This replaces it with an **EIP-1153 transient membership set**: mark the current/next/subsequent committee members once, then test each pending-exit validator in `O(1)`.

- The set is keyed to the concluding epoch (`keccak(_COMMITTEE_TSET, epoch)`), so distinct epochs never collide and the transient slots auto-clear at end of transaction - no persistent state, no clear loop.
- No raw assembly: transient access goes through OpenZeppelin 5.2's audited `SlotDerivation` + `TransientSlot` wrappers. Solc cannot express this natively - the `transient` keyword only supports value types, not mappings, even on 0.8.35.
- The old array-scanning `_isCommitteeMember` is removed; the new helpers (`_markCommitteeMembers`, `_isCommitteeMember`) encapsulate all transient access, so the call site reads as one `if (_isCommitteeMember(addr, set))`.

Binary search wasn't an option: `_eject`'s swap-and-pop can unsort a committee after a governance burn, so the test must be order-independent.

## Is it worth it? (gas sweep, worst case: exiting validators absent from every committee)

| C \ P | scan P=1 | scan P=10 | scan P=50 | scan P=100 | transient P=100 |
|---|---|---|---|---|---|
| 10  | 5,161  | 51,223  | 255,943   | 511,843   | 37,967 |
| 25  | 12,091 | 120,523 | 602,443   | 1,204,843 | 50,297 |
| 50  | 23,641 | 236,023 | 1,179,943 | 2,359,843 | 70,847 |
| 100 | 46,741 | 467,023 | 2,334,943 | **4,669,843** | **111,947** |

- **C=100, P=100: 4.67M -> 112k (~42x).** The scan alone approaches block-gas territory for a single `concludeEpoch` system call, which is a liveness risk; the transient set keeps it bounded.
- **C=100, P=10: 467k -> 85k (~5.5x).**
- **Crossover at P~2.** The only regime where the scan wins is P=1 (one validator exiting per epoch), where the transient set costs a few k more from the fixed marking overhead - negligible in absolute terms.

Verdict: worth it for any realistic exit queue, and it caps a worst case that could otherwise threaten the epoch system call.

## Trade-off to flag

This is the **first transient storage in the contract**, on the consensus path, in exchange for the bounded worst case. Transient access uses OpenZeppelin's audited `TransientSlot` / `SlotDerivation` libraries rather than hand-rolled assembly, and `deriveMapping` matches solc's native mapping slot derivation; gas is identical to the assembly version across the entire bench sweep.

## Tests

Full suite: **269 passing** (76 consensus tests confirm the exit logic is unchanged behaviorally; `CommitteeMembershipGasBench` provides the sweep above).


chasebrownn (dismissed):
Good use of transient storage. All tests pass which means functionality is preserved. Looks safe and good to merge.

chasebrownn (approved):

PR #119, opened 2026-06-08
Bump solc 0.8.26 -> 0.8.35 across the repo

Huwonk:
Pre-audit compiler modernization. Targets master so it becomes the audit baseline; the open consensus PRs (#114, #118) rebase onto it afterward.

## What

- **`foundry.toml`: `solc = "0.8.26"` -> `"0.8.35"`** (latest stable, Apr 2026). `evm_version` stays `prague`.
- Bump every exact `0.8.26` pragma (both `0.8.26` and `=0.8.26` forms) to `0.8.35` across `src/`, `test/`, `script/`, `deployments/`, `external/` (59 files). Floating `^0.8.20` / `^0.8.24` library/interface pragmas already permit 0.8.35 and are left as intentionally-reusable.
- Regenerate the 7 vendored `artifacts/*.json` under 0.8.35.

## Why now

- We're pre-audit, so there's no re-audit cost - the audit baselines on the modern compiler.
- 0.8.35 is past the **transient-storage clearing-helper codegen bug** (affected 0.8.28-0.8.33, fixed in 0.8.34), which was the reason for the conservative 0.8.26 pin. Transient storage via the high-level `transient` keyword is now usable, which lets #118 drop its `tstore`/`tload` assembly once it rebases.

## Network compatibility

- **`evm_version = prague` is unchanged** - that's the only thing the network EVM depends on. telcoin-network runs reth `v1.11.3` with a Prague EVM, which already covers EIP-1153 (transient storage) and the EIP-2537 BLS precompiles the contracts use. A compiler version is not a hardfork.
- **Uniswap genesis precompiles are unaffected**: they're bytecode-literal holders (`bytes constant = hex"..."`), so the recompile does not change their deployed genesis bytecode.
- The **core consensus contracts' bytecode does change** (new compiler), so tn-reth re-captures their genesis from the regenerated artifacts. ABI and `sol!` bindings are unchanged.

## Tests

Full suite: **263 passing** under 0.8.35. (Build warnings are forge-std's `error`-as-identifier future-keyword notices, not our code.)

## Follow-up

Once this lands: rebase #114 and #118 onto it (mechanical: pragma + artifact regen), and swap #118's transient assembly for the clean `mapping(...) transient` keyword.


grantkee (approved):
LGTM!

chasebrownn (approved):
LGTM

PR #121, opened 2026-06-10
Deploy Safe CompatibilityFallbackHandler at genesis

Huwonk:
@
## Summary

- Deploys the Safe `CompatibilityFallbackHandler` at genesis, pinned to the canonical Safe v1.4.1 address `0xfd0732Dc9E303f09fCEf3a7388Ad10A83459Ec99` - the same address it has on every EVM chain, consistent with our existing canonical pins for `SafeImpl` and `SafeProxyFactory`
- References the handler from the governance safe@'@s `setup()` call, so the genesis safe supports EIP-1271 signature validation and token callbacks out of the box
- Regenerates `precompile-config.yaml`. Note this also picks up the solc 0.8.35 bytecode changes from #119, which bumped the compiler without regenerating the yaml
- Adds `test/genesis/GenesisSafeConfig.t.sol`, which replays the genesis simulation and verifies the resulting state

## Why the canonical address matters

The canonical Safe contracts have no contract-level default fallback handler: the factory forwards opaque initializer bytes and `Safe.setup` only sets a handler when one is explicitly passed. The "automatic" behavior comes from Safe tooling (protocol-kit SDK, Safe{Wallet} UI), which injects the canonical `CompatibilityFallbackHandler` address for the chain when the caller does not specify one. Pinning the handler at that address at genesis means safes created on TN through standard tooling get it with no extra parameters.

## Test plan

- `test_fallbackHandlerDeployedAtCanonicalAddress` - handler code exists at the canonical address
- `test_governanceSafeReferencesFallbackHandler` - the governance safe@'@s `fallback_manager.handler.address` slot holds the handler, plus 3-of-7 owner config sanity checks
- `test_governanceSafeFallbackDispatch` - calls `onERC721Received` and `getMessageHash` on the governance safe address itself, proving the proxy -> singleton fallback -> handler chain end to end (the EIP-712 message hash is recomputed independently and matches)
- `test_factoryCreatedSafeUsesFallbackHandler` / `test_factoryCreatedSafeWithoutHandlerHasNone` - factory-created safes get the handler when the initializer references it, and documents that passing the zero address yields no handler
- Full suite: 268 passed, 0 failed
@


script/README.md
```diff
@@ -8,21 +8,22 @@
...
-- **SafeProxyFactory** — factory for creating Safe proxies
-- **Governance Safe** — a 3-of-6 multisig proxy configured with hardcoded owner addresses and threshold
-- **TEL supply allocation** — assigns the remaining TEL supply to `0xde1e7e`
-- **EIP-2935 / EIP-4788** — system contracts for historic block hashes and beacon block roots
-- **Multicall3** — deployed at `0xcA11bde05977b3631167028862bE2a173976CA11`
+- **Safe singleton** - the Gnosis Safe implementation contract
+- **SafeProxyFactory** - factory for creating Safe proxies
+- **CompatibilityFallbackHandler** - default Safe fallback handler (EIP-1271 signature validation, token callbacks), pinned to the canonical Safe v1.4.1 address `0xfd0732Dc9E303f09fCEf3a7388Ad10A83459Ec99` so Safe tooling that defaults the fallback handler resolves it on TN
+- **Governance Safe** - a 3-of-7 multisig proxy configured with hardcoded owner addresses and threshold, referencing the CompatibilityFallbackHandler
+- **TEL supply allocation** - assigns the remaining TEL supply to `0xde1e7e`
```
grantkee, line 15: This is okay for this PR, but we need to figure out how we will set this in genesis for mainnet using the LZ native adapter

grantkee (approved):
LGTM! Created issue #122 as a follow up

PR #123, opened 2026-06-11
Per-network deployments files with live devnet address book

Huwonk:
@
## Summary

Splits the single `deployments.json` into per-network address books resolved by chain id, and records a full live deployment on the new devnet.

| Network | Chain id | File | RPC shorthand |
| --- | --- | --- | --- |
| Testnet | `0x7e1` (2017) | `deployments/deployments-testnet.json` | `--rpc-url testnet` |
| Devnet | `0x7e1d` (32285) | `deployments/deployments-devnet.json` | `--rpc-url devnet` |
| Mainnet | TBD | `deployments/deployments-mainnet.json` | added when finalized |

## How it works

- `deployments/DeploymentsResolver.sol` maps `block.chainid` to the file. All testnet scripts and the Uniswap fork tests read and write through it, so a script pointed at a devnet RPC physically cannot touch the testnet address book. Unknown chain ids (local sims, tests) fall back to the testnet file, preserving prior behavior.
- The bash pipeline (`deploy-testnet-infra.sh`, `test-faucet-drips.sh`) applies the same rule via `cast chain-id`. An explicitly exported `TN_RPC_URL` now takes precedence over the `.env` value (previously `.env` silently overrode the environment).
- `deployments-mainnet.json` is strictly genesis-assigned (admin and Permit2 zeroed) and is the genesis source of truth: `GenerateGenesisPrecompileConfig` reads it, since genesis-assigned addresses are identical on every network.
- `deployments-devnet.json` starts each reset with genesis keys only; the deploy pipeline repopulates it.
- CREATE2 salts drop their legacy `_v2`/`_v3` suffixes. Those existed to dodge collisions with prior Adiri deployments; devnet resets target a fresh chain, and the updated comments document when a salt bump is needed again.
- `test/DeploymentsResolver.t.sol` guards the chain-id mapping and decodes all three files through the `Deployments` struct, asserting genesis keys stay byte-identical across networks - adding a key to the struct or one file but not the others fails CI.

## Devnet deployment (recorded in this PR)

[...]

PR #124, opened 2026-06-11
Add WTEL to genesis precompiles at the 0x...37E1 vanity address

Huwonk:
@
## Summary

Adds WTEL (wrapped TEL, canonical WETH9 shape) to the genesis precompile configuration so future networks - and devnet from its next reset onward - get it at the reserved vanity address:

```
0x00000000000000000000000000000000000037E1
```

The `3` reads as a sideways `w` next to the existing `7e1` = TEL convention (TEL mint precompile at `0x7e1`, governance safe at `0x7a0`), and `0x37e1` sits well clear of the EVM precompile range.

## What changed

- `GenerateGenesisPrecompileConfig` simulates `new WTEL()` and etches it at the vanity address (no constructor, no storage; name/symbol/decimals are constants, so it is a no-storage genesis entry like the SafeProxyFactory and CompatibilityFallbackHandler). `precompile-config.yaml` is regenerated with the new entry.
- `deployments-mainnet.json` records WTEL as genesis-assigned. The genesis schema test asserts it.
- **Transition state, by design:** live testnet and devnet predate this genesis entry and keep their CREATE2-deployed WTEL addresses in their respective files until their next regenesis/reset. We obviously cannot re-run genesis on running chains; the vanity address is reserved and documented now (natspec on `Deployments.WTEL`, README), and the devnet file flips to it at the next reset.
- `TestnetDeployWTEL` and pipeline step 4 are marked LEGACY: the existing has-code check already skips them on any chain whose genesis includes WTEL, so they only fire on pre-genesis chains (the live Adiri testnet). They can be retired once no such chain is operated.
- New `GenesisWTEL` tests replay the genesis simulation and exercise deposit, transfer, withdraw, and the bare-send `receive()` path against the etched code at the vanity address.

## Stacking note

This branch is cut from `feat/devnet-deployments` (#123) so the per-network deployments work rides along, per discussion. Review and merge #123 first; we will sync this branch afterward so its diff collapses to just the WTEL changes above.

## Tests

Full suite: 273 passed, 0 failed (adds 2 GenesisWTEL tests on top of #123@'@s 271).
@


chasebrownn (dismissed):
LGTM

chasebrownn (approved):

PR #126, opened 2026-06-17
consensus: generic BLS verify precompile + validator-set migration

Huwonk:
﻿## Summary

Two changes to the `ConsensusRegistry` upgrade path, on top of the compressed proof-of-possession migration:

1. **Generic BLS verify precompile.** `BlsG1` now exposes a generic `blsVerify(signature, pubkey, message)` that forwards to the native precompile, instead of a proof-of-possession-specific entrypoint, so the primitive is reusable beyond `ConsensusRegistry`. The PoP message (`intentPrefix || compressedPubkey || address`) is built on-chain by a new `BlsG1.proofOfPossessionMessage` helper and verified via `blsVerify`. It uses the compressed key and the raw address, so no on-chain decompression is needed; the genesis validators' proofs are regenerated to the new message (absorbed by the rollout).

2. **`migrateValidatorSets`.** A one-time, system-gated back-fill that reconstructs the per-status `EnumerableSet`s and the cached eligible count from `validators[].currentStatus` (the source of truth that survives a code swap, since the storage layout is a clean append over the deployed registry). Used to upgrade an existing deployment in place; idempotent and a no-op on a freshly seeded genesis.

## Validation

Full consensus + fuzz suite green (122 tests, 250 fuzz runs each), including the new `ConsensusRegistryMigrationTest` (rebuild-from-cleared, idempotent no-op, system-only access). The vendored `ConsensusRegistry.json` / `BlsG1.json` artifacts are regenerated (artifacts profile, `__$BlsG1$__` link placeholder preserved). Pairs with telcoin-network#759, whose submodule points at this commit.

## Audit (self-review)

A differential audit of this diff found 0 Critical / 0 High:

- **Medium:** `migrateValidatorSets` is an `O(totalSupply)` loop; at a large validator count a single-shot system call could exceed the gas the fork transition allots and leave the sets empty. Not triggerable at the current testnet scale (5 validators); paginate before any large-set use.
- **Low (hardened in this PR):** `proofOfPossessionMessage` packs a dynamic `bytes pubkey` between fixed fields. Added a self-contained `pubkey.length == 96` assert so the builder no longer relies on the caller's length gate.
- **Informational:** `migrateValidatorSets` is add-only (assumes empty-or-consistent sets - documented precondition); the PoP message carries no chain id or nonce, which is benign because it is a key-possession proof gated by the governance-minted ConsensusNFT and `msg.sender`.

The PoP verification is sound: the message binds the compressed pubkey (rogue-key defense), the precompile verifies under the supplied key with full signature + pubkey subgroup checks, and the registry stores the exact key it verified.

## Status

Draft for team review. Targets `master`. The associated testnet deployment approach (state-preserving relaunch vs. in-place hardfork) is under discussion; see telcoin-network#759 for the observer-sync analysis that bears on it.

## Update - BlsG1 is now an `IBlsG1` interface (supersedes the linked-library notes above)

Follow-up refactor (`a44d726`): `BlsG1` was a linked library that reached the precompile via `delegatecall`. It is now a thin `interface IBlsG1`, and `ConsensusRegistry` reaches the precompile with a direct low-level `staticcall` carrying `IBlsG1.blsVerify.selector`. This matches the protocol's existing precompile-call convention (see `StablecoinManager`), which deliberately avoids the `EXTCODESIZE` guard a typed interface call would emit against a bytecode-less precompile.

The supporting pieces moved out of `BlsG1`:
- `ProofOfPossession` and `InvalidBLSPubkey` into `IStakeManager` (an interface cannot reference a struct defined in the contract that implements it, so they sit beside `InvalidProofOfPossession`).
- `POP_INTENT_PREFIX`, `proofOfPossessionMessage`, and the `LowLevelCallFailure` guard into `ConsensusRegistry`.

The linking machinery is gone: the `foundry.toml` `libraries` pin and the `artifacts` profile are removed, and the registry bytecode now bakes `BLS_G1_ADDRESS` in directly. The note above about regenerating `BlsG1.json` with the `__$BlsG1$__` placeholder no longer applies - `BlsG1.json` is deleted, and `ConsensusRegistry.json` is regenerated with no link placeholder via a plain `forge build`. Consensus suite still 122/122.


src/consensus/ConsensusRegistry.sol
```diff
@@ -151,6 +150,36 @@ contract ConsensusRegistry is StakeManager, Pausable, Ownable, ReentrancyGuard,
...
+            address validatorAddress = ownerOf(tokenByIndex(i));
+            ValidatorInfo storage validator = validators[validatorAddress];
+            if (validator.isRetired) continue;
+
+            ValidatorStatus status = validator.currentStatus;
+            if (status == ValidatorStatus.Undefined || status == ValidatorStatus.Any) continue;
+
+            validatorSets[uint8(status)].add(validatorAddress);
+            if (_eligibleForCommitteeNextEpoch(status)) ++eligible;
+        }
```
chasebrownn, line 177: As I'm sure you know iterating total validators could be dangerous here, but I see the comment specifying its a one-time migration for existing validators on Adiri and I know there's only 5 validators right now so this should be completely fine.

chasebrownn (approved):
Honestly - everything looks good to me. The new solution is clean. I hate low-level calls, but there's really no healthy alternative and we can exit gracefully with the conditional you implemented if it fails. Precompiles dont have revert statements which is what we lose if we used a low-level call on a solidity contract. So in this case it's acceptable. The solution looks well covered by tests. Good work.

src/consensus/ConsensusRegistry.sol
Huwonk: yeah, after we get testnet upgraded, this function will get removed ideally.

grantkee (approved):
LGTM! Thank you for taking this on. I know it's a bit odd having the BLS library in a different repo. I'm thinking through how we can make that a little less awkward for testing, but this is a great step

src/consensus/ConsensusRegistry.sol
grantkee: Added issue #128 to track this post-migration

PR #131, opened 2026-07-13
fix(ci): materialize ring buffer tail on GitAttestationRegistry grow

### LINKED ISSUE #154: Cantina finding 13: Grow path under-allocates ringBuffer
Cantina review finding 13 (Low).

`GitAttestationRegistry.setBufferSize` corrupted the ring buffer on a grow: the rebuild pushed only the old records while setting the new size, leaving the head equal to the array length so the next attestation reverted out of bounds.

Remediation: #131 pads the buffer to the new size so storage length always equals `bufferSize`. This issue closes automatically when that PR merges.

Huwonk:
## Finding

Cantina review (Cryptara). `setBufferSize` corrupts the ring buffer when growing.

On a grow (`newSize > bufferSize`), `itemsToCopy = min(old, new) = old`. The rebuild deletes `ringBuffer` and pushes only `itemsToCopy` records, while setting `bufferSize = newSize` and `head = old % newSize = old`. That leaves `head == ringBuffer.length`:

- the next `attestGitCommitHash` writes `ringBuffer[head]` out of bounds -> `Panic(0x32)`
- `gitCommitHashAttested` reverts for any hash not in the first `old` slots
- a further grow or same-size call reverts

The only recovery is a counter-intuitive shrink to a size <= the old length. This is reachable by an ordinary admin capacity increase, so the tooling can brick itself. CI tooling only, no funds at risk.

## Fix

- Pad the buffer to `newSize` with empty records after the copy, so storage length always equals `bufferSize`.
- Assert `ringBuffer.length == bufferSize` and `head < bufferSize`.

## Tests

Four regression tests that fail against the prior implementation and pass here: attest after grow, missing lookup after grow, full-slot materialization, and consecutive grows. The pre-existing `testResizeBuffer` missed the bug because it never attests after growing.

Full suite: `test/CI` 9 passed.

Closes #154.

### DIFF
--- a/src/CI/GitAttestationRegistry.sol
+++ b/src/CI/GitAttestationRegistry.sol
@@ -69,9 +69,18 @@ contract GitAttestationRegistry is AccessControl {
         for (uint8 i; i < itemsToCopy; ++i) {
             ringBuffer.push(newBuffer[i]);
         }
+        // materialize the remaining slots so storage length always equals `bufferSize`; on grow
+        // `itemsToCopy` is the smaller size, so [itemsToCopy, newSize) must exist before `head` reaches them
+        for (uint8 i = itemsToCopy; i < newSize; ++i) {
+            ringBuffer.push(GitCommitHashRecord(bytes20(0x0), false));
+        }
 
         bufferSize = newSize;
         head = itemsToCopy % newSize;
+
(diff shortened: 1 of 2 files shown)


grantkee (approved):
LGTM!

chasebrownn (approved):

PR #135, opened 2026-07-14
Add governance-ejection lifecycle test coverage

grantkee:
- Adds `ConsensusRegistryEjectionTest.t.sol`, covering `burn` and slash-to-zero ejection through `_ejectFromCommittees`/`_eject`: epoch-conclusion continuity over shrunken committees, epoch-info ring-buffer consistency, stake confiscation accounting (including the initial-stake/rewards split), retired-skip behavior in `applyIncentives`/`applySlashes` batches, and guard/atomicity checks (emptying the committee, double-burn, unseated PendingActivation burns, paused-state burns).
- Pins the exact swap-and-pop order `_eject` produces on the stored committee arrays, since the Rust epoch-record producer reads them with order-sensitive positional comparisons.
- Extends `ConsensusRegistryTestUtils.sol` with shared helpers: `_sortedGenesisCommittee`, `_seatCommittees`, `_addFifthValidator`, `_assertCommitteeExcludes`.
- No production contract changes - test-only.

Relates to #143 (test coverage pinning the ejection storage order).

### DIFF
--- a/test/consensus/ConsensusRegistryEjectionTest.t.sol
+++ b/test/consensus/ConsensusRegistryEjectionTest.t.sol
@@ -1,0 +1,760 @@
+// SPDX-License-Identifier: MIT or Apache-2.0
+pragma solidity 0.8.35;
+
+import "forge-std/Test.sol";
+import { ConsensusRegistry } from "src/consensus/ConsensusRegistry.sol";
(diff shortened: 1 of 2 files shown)


Huwonk (approved):
LGTM

PR #136, opened 2026-07-15
fix(consensus): slash settlement, stake top-ups, issuance withdrawal, emergency exit

Huwonk:
## Summary

Five changes from the audit-findings list, plus settlement-accounting fixes that surfaced while wiring them together:

- **Slash-aware reward weights**: `applyIncentives` now derives each validator's weight from its effective stake - the version's stake amount, capped to the outstanding balance when slashes have reduced it below that amount. Rewards above the stake amount never add weight.
- **Slashed-stake top-ups**: new `topUpSlashedStake` restores a slashed validator's balance ledger to the full stake amount. The exact deficit must be provided as `msg.value`. A `topUpAuthorityRequired` toggle (governance-settable via `setTopUpAuthorityRequired`) restricts top-ups to governance; while it is off, the validator or its delegator may self-serve. Governance may always top up.
- **Issuance withdrawal**: new `Issuance.withdraw` plus a governance-gated `issuanceWithdrawal` on the registry, allowing recovery of consolidated slash funds and prior issuance allocations.
- **Emergency exit**: `unstake` takes an `emergencyExit` flag that permanently forfeits accrued rewards and returns only the stake balance, so a validator can still exit when the Issuance balance cannot cover the rewards owed. The `RewardsClaimed` amount reflects the forfeiture.
- **No orphaned slash funds**: `_consensusBurn` now consolidates the validator's entire stake-backed native TEL on Issuance. Previously it moved only `min(outstandingBalance, initialStake)`, so burning a partially slashed validator stranded the slashed remainder on the registry permanently.

## Design note: when slashed native TEL moves

The registry holds exactly `stakeAmount` native TEL per staked validator; reward balances are ledger entries backed by the Issuance pool. We kept partial slashes as ledger-only decrements and instead consolidate the confiscated native on Issuance at settlement:

- top-up: the deficit `msg.value` is forwarded straight to Issuance (the registry already holds the full stake-backed native)
- unstake: the existing `stakeAmt - bal` consolidation
- burn / slash-to-zero: the full `initialStakeAmt` (the orphan fix above)

An earlier draft moved `slash.amount` of native to Issuance inside `applySlashes` itself. That double-counts against the settlement paths (`_unstake` and the `upgradeValidatorStakeVersion` downgrade path would consolidate the same funds again, draining native belonging to other validators) and over-sends when a slash overlaps accrued rewards, whose native lives on Issuance rather than the registry. Settlement-time consolidation preserves the single invariant every money path already assumes, and keeps `applySlashes` free of per-slash external calls on the system-call hot path. The only timing difference is when the confiscated TEL becomes part of the reward pool, and governance can always rebalance via `allocateIssuance` / `issuanceWithdrawal`.

`topUpSlashedStake` access control is enforced in the direction the flag name implies: with `topUpAuthorityRequired` on, only governance may top up; with it off, the stake originator (validator or delegator) may as well. Third parties can never top up.

## Comment/doc updates

- `Issuance` and `allocateIssuance` natspec no longer claim allocated TEL is unrecoverable (superseded by `issuanceWithdrawal`)
- completed/added natspec for the new errors, events, and functions in both interfaces
- the `applyIncentives` weight comment now describes the balance-capped derivation

## Testing

25 new tests covering: partial-slash ledger semantics and settlement consolidation, full-slash and partial-then-full burn fund flows, burn of a slashed validator (orphan regression), slash-reduced reward weights and the stake-amount weight cap, top-up happy paths (self, owner-with-flag, delegator), the full slash -> top-up -> unstake round trip, all top-up revert paths (authority flag, non-recipient, not slashed, wrong value, bad status, unknown validator, paused), the authority toggle, issuance withdrawal (happy, non-owner, insufficient balance, direct-call gate), and emergency exit (reward forfeiture, and exit with a drained reward pool where normal unstake reverts).

Full suite: 243 passed, 0 failed (fast profile).

## Merge-order note

PR #134 touches the same regions (`IStakeManager.unstake`, the `delegateStake` surroundings, and the delegated test fixtures) and will need a small rebase whichever lands second. PRs #132, #133, and #135 do not overlap; #135's new ejection tests compile unchanged against this branch's `unstake` signature since they never call it.

### DIFF
--- a/src/consensus/ConsensusRegistry.sol
+++ b/src/consensus/ConsensusRegistry.sol
@@ -51,6 +51,10 @@ contract ConsensusRegistry is StakeManager, Pausable, Ownable, ReentrancyGuard, 
     /// than three set-length reads. Invariant-tested against the eligible set sizes.
     uint256 internal eligibleValidatorCount;
 
+    /// @notice When true, `topUpSlashedStake` is restricted to governance; when false, validators
+    /// and their delegators may also restore their own slashed stake
+    bool public topUpAuthorityRequired;
+
     /// @dev Signals a validator's pending status until activation/exit to correctly apply incentives
     uint32 internal constant PENDING_EPOCH = type(uint32).max;
 
@@ -101,8 +105,12 @@ contract ConsensusRegistry is StakeManager, Pausable, Ownable, ReentrancyGuard, 
             if (isRetired(reward.validatorAddress)) continue;
 
             uint8 rewardeeVersion = validators[reward.validatorAddress].stakeVersion;
-            // derive validator's weight using initial stake for stability
-            uint256 stakeAmount = versions[rewardeeVersion].stakeAmount;
+            // derive validator's weight from its effective stake: the version's full stake amount,
+            // capped to the outstanding balance when slashes have reduced it below that amount
+            // (restorable via `topUpSlashedStake`). Rewards above the stake amount never add weight
+            uint256 versionStakeAmount = versions[rewardeeVersion].stakeAmount;
+            uint256 balance = balances[reward.validatorAddress];
+            uint256 stakeAmount = balance < versionStakeAmount ? balance : versionStakeAmount;
             uint256 weight = stakeAmount * reward.consensusHeaderCount;
 
             totalWeight += weight;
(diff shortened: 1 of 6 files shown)


Huwonk:
Closing in favor of a split: per our review process each Medium finding gets a standalone PR, and this draft bundled two Mediums together with feature work. Replacements: #137 (slashed stake stranded after burn, Medium), #138 (reward underfunding blocks principal unstaking, Medium), #139 (stake top-ups, governance issuance withdrawal, balance-capped incentives). The branch remains for reference.

PR #138, opened 2026-07-20
fix(consensus): emergency exit path so reward underfunding cannot block unstaking

### LINKED ISSUE #144: Cantina finding 3: Reward underfunding can block principal unstaking
Cantina review finding 3 (Medium).

`Issuance.distributeStakeReward` reverts when the Issuance balance cannot cover the rewards owed, and `unstake` paid principal and rewards through a single call, so reward underfunding blocked reclaiming staked principal.

Remediation: #138 adds an `emergencyExit` flag to `unstake` that permanently forfeits accrued rewards and returns the principal alone; the normal path is unchanged. This issue closes automatically when that PR merges.

Huwonk:
## Finding

Cantina review, Medium. Reward underfunding can block principal unstaking.

`unstake` settles stake and accrued rewards in a single `Issuance.distributeStakeReward` call, which reverts `InsufficientBalance` whenever the Issuance contract's balance cannot cover `rewardAmount + msg.value`. The stake portion is forwarded as `msg.value` and is always fully backed by the registry, but the reward portion depends on the Issuance pool being sufficiently funded by governance allocations. An underfunded pool therefore blocks withdrawal of the principal too: a fully exited validator has no path to reclaim its stake until governance tops the pool back up.

## Fix

- `unstake(address validatorAddress, bool acceptRewardShortfall)`: when the flag is set, the rewards leg is capped at the Issuance contract's available balance, so principal withdrawal never depends on the reward pool's funding level.
  - With a funded pool the shortfall-accepting path is identical to a normal unstake: nothing payable is ever forfeited. This also keeps one party from burning the other's payable rewards on a delegated stake, since either the validator or its delegator may call `unstake`.
  - On a genuine shortfall the recipient receives the stake plus the payable portion of the rewards; only the unpayable remainder is forfeited. Funding-wise this path cannot fail.
- The capping happens in `_unstake` before the Issuance call, so `Issuance.distributeStakeReward` keeps its two-parameter signature.
- The normal path (`acceptRewardShortfall = false`) still reverts on underfunding, preserving the option to wait for a governance refill instead of forfeiting.
- The claim path and the version-change refund path are unchanged; the burn path requests shortfall semantics (inert there, since the balance ledger is already settled).

Note: this changes the `unstake` selector, so downstream callers and any off-chain bindings need the new boolean argument.

## Tests

- `test_unstake_acceptRewardShortfall_paysFullRewardsWhenFunded`: with a funded pool, the flag pays stake plus full rewards, identical to a normal unstake.
- `test_unstake_acceptRewardShortfall_insufficientIssuanceBalance`: with the reward pool emptied, a normal unstake reverts `InsufficientBalance` while the shortfall-accepting path still returns the stake, forfeiting only the unpayable rewards. Fails against the prior implementation, which had no way through.
- `test_unstake_acceptRewardShortfall_partialShortfall`: with the pool covering only part of the accrued rewards, a normal unstake still reverts while the shortfall-accepting path pays the stake plus the payable portion and drains the pool to zero.
- Existing unstake call sites updated to the two-argument form.

Full suite: 221 passed, 0 failed, 2 skipped.

Closes #144.

### DIFF
--- a/src/consensus/ConsensusRegistry.sol
+++ b/src/consensus/ConsensusRegistry.sol
@@ -552,7 +552,7 @@ contract ConsensusRegistry is StakeManager, Pausable, Ownable, ReentrancyGuard, 
             if (refundAmount > 0) {
                 balances[validatorAddress] -= refundAmount;
                 // Route through Issuance (same pattern as _unstake)
-                Issuance(issuance).distributeStakeReward{ value: refundAmount }(recipient, 0, false);
+                Issuance(issuance).distributeStakeReward{ value: refundAmount }(recipient, 0);
             }
 
             // consolidate confiscated slash remainder on Issuance (same as _unstake pattern)
@@ -610,7 +610,7 @@ contract ConsensusRegistry is StakeManager, Pausable, Ownable, ReentrancyGuard, 
         // permanently retire the validator and burn the ConsensusNFT
         _retire(validator);
 
-        // return stake and send any outstanding rewards unless forfeited via emergency exit
+        // return stake plus rewards; an emergency exit forfeits only rewards Issuance cannot cover
         uint256 stakeAndRewards = _unstake(validatorAddress, recipient, emergencyExit);
 
         emit RewardsClaimed(recipient, stakeAndRewards);
--- a/src/consensus/ConsensusRegistry.sol
+++ b/src/consensus/ConsensusRegistry.sol
@@ -552,7 +552,7 @@ contract ConsensusRegistry is StakeManager, Pausable, Ownable, ReentrancyGuard, 
             if (refundAmount > 0) {
                 balances[validatorAddress] -= refundAmount;
                 // Route through Issuance (same pattern as _unstake)
-                Issuance(issuance).distributeStakeReward{ value: refundAmount }(recipient, 0);
+                Issuance(issuance).distributeStakeReward{ value: refundAmount }(recipient, 0, false);
             }
 
             // consolidate confiscated slash remainder on Issuance (same as _unstake pattern)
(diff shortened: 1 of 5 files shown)


src/consensus/ConsensusRegistry.sol
```diff
@@ -599,7 +599,7 @@ contract ConsensusRegistry is StakeManager, Pausable, Ownable, ReentrancyGuard,
     }
 
     /// @inheritdoc StakeManager
-    function unstake(address validatorAddress) external override whenNotPaused nonReentrant {
+    function unstake(address validatorAddress, bool emergencyExit) external override whenNotPaused nonReentrant {
```
grantkee: nit: can we make `emergencyExit` more descriptive? Maybe `exitWithoutRewards` or `abandoneRewards` or `forfeitRewards` ?

grantkee (dismissed):
LGTM - one nit about using the var name `emergencyExit` since it isn't immediately obvious unstaking with this bool would result in reward loss

Huwonk:
Agreed on the naming. One wrinkle: since the latest commit the flag no longer forfeits all rewards - it caps the rewards leg at the Issuance balance, so a funded pool pays in full and only a genuine shortfall is forfeited. That makes exitWithoutRewards / forfeitRewards slightly overstate the loss, so we went with acceptRewardShortfall: it names exactly what the caller consents to. Renamed in 6ac57fb along with the test names and natspec.

grantkee (dismissed):

chasebrownn (dismissed):

Huwonk:
Conflicts with master resolved in d75891d (union of the slashing-settlement tests from #137 and the reward-shortfall tests here; the slashed-then-unstake test gains the new bool argument). Heads up on one extra hunk in the merge commit: test_applySlashes_exactBalanceBoundary asserted the pre-#137 confiscation amount (+1 wei) and #137 did not update it, so that test currently fails on master itself. The merge aligns its expectation with the full-initial-stake consolidation; if a standalone hotfix to master is preferred instead, happy to split it out.

grantkee (approved):

PR #140, opened 2026-07-20
fix(consensus): settle stake version changes automatically at epoch boundaries

### LINKED ISSUE #146: Cantina finding 5: Validators can adopt stake versions before epoch activation
Cantina review finding 5 (Low).

`upgradeValidatorStakeVersion` bounds the target by the global latest version rather than the epoch-active one, so validators can adopt a version before it activates at the next epoch start.

Disposition: accepted. Early adoption only targets configurations governance has already authored: a stake-increasing adoption escrows the exact deficit immediately, and a stake-decreasing adoption by a serving validator is governed by the delayed two-step downgrade introduced in #140, so no collateral can leave ahead of the version's activation. Forcing adopters to additionally wait for the epoch flip would add friction without a protective benefit.

### LINKED ISSUE #147: Cantina finding 6: Pending stake versions can retroactively alter epoch reward weights
Cantina review finding 6 (Low).

`applyIncentives` weighs an epoch's consensus headers by each validator's stake version at epoch end, so a mid-epoch version adoption changes the weight applied to headers already produced.

Disposition: accepted. A stake-increasing adoption escrows the exact deficit up front, so the added weight is fully collateralized, and the incentives calculation caps the weight basis at the validator's actual balance (#139). Stake decreases by serving validators move no value and change no version until finalized in a later epoch (#140). A complete fix would require per-epoch snapshots of every validator's stake version, which is disproportionate for a Low.

### LINKED ISSUE #156: Cantina finding 15: Active validators can shed slashable stake via version downgrade
Cantina review finding 15 (Medium).

A committee-serving validator could adopt a governance-authored lower-stake version and receive the surplus refund immediately, with no exit delay, shrinking its slashable collateral and the burn-confiscation bound ahead of an anticipated slash.

Remediation: #140 turns in-service downgrades into a two-step request and finalize flow that can only finalize after the next epoch boundary, with the refund computed from finalize-time balances so interim slashes shrink it. Staked validators keep the immediate path since a pre-activation validator can already unstake in full at any time. This issue closes automatically when that PR merges.

Huwonk:
## Findings

Cantina review, Medium. Active validators can shed slashable stake via version downgrade (finding 15), plus the two accepted Low findings this design now resolves outright: validators can adopt stake versions before epoch activation (finding 5) and pending stake versions can retroactively alter epoch reward weights (finding 6).

The prior remediation on this branch (a two-step downgrade finalizable after one epoch boundary) left validator-controlled settlement timing in place: a pending downgrade could be held open indefinitely and finalized ahead of a detected slash in any later epoch. Review feedback on that approach was correct, and the ideal it pointed to - applying all changes at epoch boundaries - is what this implements.

### One boundary system call

`concludeEpoch(address[] newCommittee, RewardInfo[] rewardInfos, Slash[] slashes)` becomes the single epoch-boundary system call; the separate `applyIncentives`/`applySlashes` externals are removed. The internal stage order is a security invariant enforced by the contract rather than an ordering obligation on the protocol client:

1. incentives, weighted by the versions active during the closing epoch
2. slashes, landing on full old-version collateral (slash-to-zero ejections included; the committee-length check runs after this stage so an ejection validates the incoming committee against the post-ejection size)
3. version-queue settlement, computing refunds from post-slash balances
4. epoch rotation (existing activation/exit processing)
5. refund delivery, last, after every ledger and epoch mutation

No value can leave the registry at a boundary ahead of that boundary's slashes, by construction.

### The version-change queue

`requestStakeVersionChange` replaces `upgradeValidatorStakeVersion`. For in-service validators (PendingActivation/Active) it only queues; the change applies automatically inside `concludeEpoch`. Validators have no control over settlement timing, so no request timing beats a detected slash - and since versions flip only at boundaries, reward weights are stable within an epoch by construction, resolving findings 5 and 6 as a side effect.

- Increases escrow the exact deficit in the queue entry - never in `balances`, where it would be immediately claimable as fake rewards (`rewards = balance - stakeAmount`). The escrow credit and the version flip apply atomically inside the system call, so no state ever shows the escrow against the old reward floor.
- Decreases move no value at request time and age `STAKE_DECREASE_DELAY_EPOCHS = 1` extra boundary, covering slashes whose detection lags the offense by an epoch. Settlement uses the existing tiered math from post-slash balances; the slashed remainder of the surplus consolidates on Issuance.
- A repeat request overwrites the entry, returning the prior escrow to its recorded funder (never the reward recipient, so neither party of a delegation can harvest the other's escrow); `cancelStakeVersionChange` withdraws it.
- Retirement (unstake, governance burn, slash-to-zero) drops the entry and credits the escrow back to the funder - it never became stake and is not confiscable.
- `Staked` validators keep the immediate lane: not in service, never in a committee, and able to unstake in full at any time.

### Boundary hardening

- Refunds are pushed through Issuance with a bounded gas stipend (`REFUND_GAS_LIMIT`); a failed push credits `claimableRefunds` for a later `claimRefund` pull. No recipient can revert or gas-grief `concludeEpoch`.
- Queue processing never reverts: stale entries are skipped and cleared, unaged decreases stay queued, and iteration is bounded by the validator count (one ConsensusNFT-gated entry per validator).
- Credits are keyed by recipient and survive retirement; `concludeEpoch` remains non-pausable while request/cancel/claim are pause-gated.
- Storage is appended at the ConsensusRegistry tail (queue set, entries, credits), preserving every existing slot for in-place upgrades.

## Protocol companion

The `concludeEpoch` ABI change is breaking in both directions, so this lands in lockstep with the telcoin-network client change (single system call carrying the rewards counter output and an empty slashes array until slashing ships): Telcoin-Association/telcoin-network#1012. When programmatic slashing ships, the protocol only fills the `slashes` argument - the ordering guarantee lives in the contract.

## Tests

- Reworked existing suites to the unified boundary call (a `_concludeEpoch` family of helpers wraps committee/rewards/slashes variants).
- New `ConsensusRegistryVersionQueueTest` suite: request/cancel/claim and overwrite flows, escrow accounting and funder semantics on delegated stakes, burn/exit-while-queued lifecycle, a hostile recipient mock proving a reverting or gas-burning recipient cannot disrupt the boundary (settlement falls back to a pull credit), escrow invisibility to reward claims before and after the flip, mixed-direction multi-validator boundaries, and registry native-balance conservation across the full flow.
- The finding-15 core scenario is pinned directly: a decrease queued while unslashed settles only after a boundary slash lands, refunding from the post-slash balance.
- Full suite: 256 passed, 0 failed, 2 skipped (250-run fuzz profile).

Closes #156. Closes #146. Closes #147.


Huwonk:
Rebased over master (now carrying #131-#141) in merge commit cce84f8. The incoming test sections (slashing settlement, reward-shortfall unstaking, top-ups, capped weights, epoch projection) are ported to the unified concludeEpoch helpers, with a survivor-committee variant for slash-to-zero boundaries; #139's balance-capped incentive weights folded into _applyIncentives untouched. One comment touch-up: the slash-consolidation note in _applySlashes now also names the queued stake-decrease settlement as a consolidation point, which this PR introduces. Full suite green at both fuzz profiles (291 passed). Reminder for merge timing: this lands in lockstep with the client change in Telcoin-Association/telcoin-network#1012 (breaking concludeEpoch ABI), which then needs its submodule pin and artifact regeneration before e2e.

chasebrownn (dismissed):
build and tests run cleanly

Huwonk:
Gas hardening follow-up in 44d0e5a: boundary settlement refunds now accrue as claimRefund credits instead of being pushed, so concludeEpoch performs no recipient-facing external calls at all and its cost is storage-bounded regardless of recipient behavior. Recipients withdraw with one claimRefund call; user-initiated escrow returns (cancel/overwrite) keep the gas-capped push with credit fallback. Also adds ConcludeEpochGasBench measuring worst-case boundaries against the client's 30M system-call budget: N=100 committee + 100 rewards + 10 slashes + 100 settling decreases = 7.66M (was 9.48M with pushes, plus up to 100k per hostile recipient that no longer exists as a vector); N=150 = 11.5M. Full suite green at both fuzz profiles (295 passed).

chasebrownn (approved):

grantkee (approved):
LGTM - good work @Huwonk

PR #161, opened 2026-07-29
feat(consensus): system-call worker config updates and owner-raised strategy ceiling

Huwonk:
WorkerConfigs gains protocol-facing runtime updates and a governable strategy ceiling.

## System-call mutators

`setWorkerConfigsData(workerIds, datas)` and `setWorkerConfigsValue(workerIds, values)` are `onlySystemCall` batch setters letting the protocol adjust per-worker fee data and values at runtime. Each preserves the worker's other fields, and both are gated to previously configured workers: an unconfigured id reverts `MissingWorkerConfig`, so a system call can never mark a worker as configured and fabricate the coverage `setNumWorkers` relies on for safe worker-set expansion. Writes are single-field through a storage pointer.

## Owner-raised MAX_STRATEGY

The strategy ceiling moves from a compile-time constant to storage behind the unchanged `MAX_STRATEGY()` getter, raised by `setMaxStrategy` (owner-only, strictly increasing - lowering would strand stored configs above the new bound, so a non-increasing value reverts `InvalidStrategy`). The variable is appended at the storage tail so every pre-existing slot of the deployed contract is preserved for an in-place upgrade; the interface getter mutability relaxes `pure` to `view`.

## WorkerConfig.data widens to uint184

`uint8 + uint64 + uint184` packs the struct into exactly one storage slot (the prior uint128 left 56 bits unused). Constructor, setters, getters, and the `WorkerConfigUpdated` event all carry the widened type.

## Protocol lockstep

The `sol!` bindings in telcoin-network (`WorkerConfigs::constructorCall` in the pre-genesis deploy and `getAllWorkerConfigsCall` in the per-epoch fee read) encode `uint128` datas today and must move to `uint184` together with this change; the wiring that will emit the new system calls lands with whichever protocol feature drives runtime fee updates. Both fold naturally into the pending client-side pin update.

## Tests

Twelve new tests: batch data/value updates preserving sibling fields with event assertions, `OnlySystemCall` rejection of owner and arbitrary callers, `LengthMismatch`, `MissingWorkerConfig` on unconfigured ids plus a regression pinning that a system call cannot fabricate `setNumWorkers` coverage, ceiling raises unlocking new strategy ids with old/new event values, non-owner and non-increasing `setMaxStrategy` rejections, and `uint184` max-value roundtrips. Existing suite swept to the widened type.

Full suite: 313 passed, 0 failed, 2 skipped.


grantkee (approved):
LGTM! Nice and straightforward - thanks @Huwonk

PR #162, opened 2026-07-30
refactor(consensus): split applyIncentives and applySlashes back out of concludeEpoch

Huwonk:
Splits `applyIncentives` and `applySlashes` back out of the unified `concludeEpoch`, resolving two latent consensus-safety gaps a review of telcoin-network#1012 surfaced in the three-argument design.

## The gaps

The unified `concludeEpoch(newCommittee, rewardInfos, slashes)` validated `newCommittee.length` against `nextCommitteeSize` *after* applying slashes inside the same call, and `_ejectFromCommittees` scrubbed the stored committees. But the protocol client can only read the size and assemble the committee *before* the call. Once slashing ships:

- a slash-to-zero ejection shrinks `nextCommitteeSize` mid-call, so the length check reverts against a committee the client built from the pre-slash view - and a revert in the epoch close is a deterministic fleet-wide halt (every node replays the same certified output and hits the same revert);
- `_ejectFromCommittees` cannot touch the caller's calldata committee array, so a validator ejected mid-call can still be seated in the committee two epochs out, silently eroding BFT quorum margin.

## The split

`applyIncentives(RewardInfo[])` and `applySlashes(Slash[])` return as separate `onlySystemCall` externals; `concludeEpoch` reduces to `concludeEpoch(address[] newCommittee)`, keeping the version-queue settlement, the length/size validation, epoch rotation, and validator-queue processing. The protocol sequences the closing block as:

1. `applyIncentives(rewardInfos)` - weights on pre-slash collateral
2. `applySlashes(slashes)` - lands on full old-version collateral, ejections included
3. read post-slash `getNextCommitteeSize()` and the eligible pool, assemble the committee
4. `concludeEpoch(newCommittee)` - settles the version queue from post-slash balances, rotates the epoch

The economic stage order is preserved; the committee reads are now post-slash *by construction*, so both findings dissolve. The rewards-then-slashes-then-settlement ordering is now a caller obligation rather than an on-chain atomic guarantee - documented in the natspec, `design.md`, and `invariants.md`.

Restored selectors are byte-identical to the pre-fork ABI (`concludeEpoch(address[])` = `0x7b55a300`, `applyIncentives((address,uint256)[])` = `0x0a36cdef`), so the client's pre-fork and post-fork call shapes converge and the replay gate simplifies. The unified three-argument selector never executed on a live chain, so there is no replay debt in changing it.

## Sentinel-address hardening

A `0xFF...FF` genesis validator currently mints, activates, and bricks the first boundary it participates in: `isRetired` -> `_exists` reverts on the `type(uint160).max` sentinel token id instead of returning false. The genesis constructor now rejects both sentinel addresses (`0` was already rejected, `type(uint160).max` is added), and the reward and slash loops skip both defensively so a malformed system-call entry cannot stall the boundary.

## Gas benchmark

`ConcludeEpochGasBench` now measures each boundary call separately against the client's per-call budget (`SYSTEM_CALL_GAS_BUDGET = 100_000_000`, mirroring `SYSTEM_CALL_GAS_LIMIT` in telcoin-network's `crates/tn-reth/src/evm/mod.rs`). The two former log-only cases are real assertions. Measured worst cases under the split: `applyIncentives` ~356k at N=100, `applySlashes` negligible, `concludeEpoch` 7.30M at N=100 with a 100-entry settlement wave and 33.88M at a 1000-entry wave. The bench natspec records the pre-split figures against the earlier 30M budget (7.66M at N=100, 34.26M at 1000 entries, ~860-entry ceiling) since that history is otherwise unreachable from the pinned submodule.

## Tests

Full suite green: 313 passed, 0 failed, 2 skipped. The `_concludeEpochWith*` helpers now run the real two-call sequence (`applyIncentives`/`applySlashes` then `concludeEpoch`); the atomicity test targets `applySlashes` directly since a reverted slash batch no longer rolls back the conclude call.

Artifacts regenerated (`make update-artifacts`); only `ConsensusRegistry.json` changes.

## Companion

Pairs with the telcoin-network client rework (the `feat/unified-conclude-epoch` branch, PR #1012) which must (a) make three sequential system calls instead of one, and (b) raise the per-call gas budget - see the PR discussion for the required `SYSTEM_CALL_GAS_LIMIT` extraction.


Huwonk:
Client companion updated: telcoin-network#1012 now makes the three sequential system calls (applyIncentives -> applySlashes -> concludeEpoch(address[])) with the committee shuffle running after applySlashes commits, so the eligible-pool read and the committee reflect any ejections before the size check. cargo check passes with --features adiri (exercises the legacy fork-replay path too). The 100M per-call budget was already on that branch (commit 6897f761), so the split's gas requirement is satisfied. #1012 stays draft until this merges and the submodule pin + artifacts advance.

Huwonk:
Folded in the settlement gas-bound documentation the auditor asked for (commit 3fbfb22), since it depends on this PR's bench and touches the same design.md/invariants.md the split reworks - a separate PR would have conflicted or stacked. It records the per-boundary version-queue ceiling (~30k/entry, ~33.9M for a 1000-entry wave vs the 100M budget, ~2,900-entry headroom; ~860 under the old 30M budget where 700 entries measured ~25.4M), notes it is governance-gated, and documents the two mitigations - pagination or the lazy _effectiveStakeVersion resolution - for if the validator ceiling ever approaches it.

grantkee (approved):
LGTM

PR #165, opened 2026-08-11
fix(consensus): hold delegated EOAs to their key in delegateStake

### LINKED ISSUE #163: delegateStake silently switches from ecrecover to ERC-1271
**Severity: HIGH.** Verified verbatim against both sources.

`ConsensusRegistry.delegateStake` verifies the validator's EIP-712 signature with
`SignatureCheckerLib.isValidSignatureNowCalldata(validatorAddress, digest, validatorEIP712Signature)`
(`src/consensus/ConsensusRegistry.sol:623`). Solady branches on code size first:

```solidity
if iszero(extcodesize(signer)) {
```

(`lib/solady/src/utils/SignatureCheckerLib.sol:90`). A 7702-delegated EOA carries 23 bytes of designator code, `0xef0100 || address`. `EXTCODESIZE` returns 23, the branch is not taken, and `ecrecover` is never reached. The check becomes an ERC-1271 `isValidSignature` staticcall against whatever contract the EOA delegated to.

Two consequences.

**Liveness:** a validator using any 7702 smart-account wallet that does not implement ERC-1271 with matching semantics cannot be onboarded through `delegateStake`. Their valid ECDSA signature is ignored and the call reverts `NotValidator`. Nothing in the contract, the docs, or the error message tells them why.

**Authorization:** if the delegate's `isValidSignature` is permissive, a two-key authorization degrades to one. The digest is public (`tn_delegationDigest` exposes it), so anyone who can get a permissive delegate to approve it can bind an arbitrary delegator address. `_getRecipient` (`src/consensus/StakeManager.sol:283-289`) returns the delegator, so that address receives all `claimStakeRewards` and `unstake` proceeds.

Guardrails narrow this considerably. The attacker still needs the validator's BLS key to pass `_verifyProofOfPossession` (`:603`), the full stake amount in `msg.value` (`:607`), and the validator must be in `Undefined` status (`:611`). That is a small population, but the failure mode is total loss of stake and rewards for anyone in it.

## Remediation

Call `ECDSA.recoverCalldata` directly instead of `SignatureCheckerLib`, or check `_isDelegationDesignator(validatorAddress.code)` and route explicitly. The design intent here is plainly "prove you hold the validator key," and a code-size probe is the wrong way to test that once EOAs can have code.

### LINKED ISSUE #164: 7702 dissolves EOA screening as a durable mitigation
**Severity: MEDIUM.**

`ConsensusRegistry.mint` (`src/consensus/ConsensusRegistry.sol:902-910`) has no EOA check and correctly uses `_mint`, not `_safeMint`, so there is no `onERC721Received` callback to abuse at whitelist time. Governance can screen a prospective validator as an EOA before minting.

That screening no longer holds. Post-7702 the validator can attach a delegation after the mint, pointing at code whose fallback reverts. The downstream consequence is `burn` (`:913`), which routes through `_consensusBurn` (`:1267`) to `StakeManager._unstake` (`src/consensus/StakeManager.sol:209`) to `Issuance.distributeStakeReward` (`src/consensus/Issuance.sol:27`), and that function pushes value with a full-gas `recipient.call` and reverts the whole transaction on failure (`src/consensus/Issuance.sol:34-35`).

A validator with a reverting delegation makes `burn` permanently uncallable against itself. `pause()` does not cover it either, since the ConsensusNFT functions are explicitly outside the pause scope (`ConsensusRegistry.sol:1449`, with the exclusion stated in the doc comment above it). `burn` also carries no `nonReentrant`.

## Remediation

Convert the `burn` payout to a pull-based credit, the same pattern `_processStakeVersionQueue` already uses.

Huwonk:
Closes #163, closes #164.

## #163 - `delegateStake` silently switches from ecrecover to ERC-1271

Confirmed, both halves, with tests that fail on `master`.

`SignatureCheckerLib.isValidSignatureNowCalldata` branches on `extcodesize` before anything else. A 7702-delegated EOA carries 23 bytes of `0xef0100 || implementation`, so the branch is not taken and `ecrecover` is never reached. The check becomes an ERC-1271 staticcall against whatever the EOA currently delegates to.

- **Liveness.** `test_delegateStake_delegatedEOA_acceptsKeySignature` reverts `NotValidator` on `master` even though the validator signed the digest with its own key. Any 7702 wallet without a matching ERC-1271 handler is unonboardable, with nothing in the revert to say why.
- **Authorization.** `testRevert_delegateStake_permissiveDelegateCannotAuthorize` succeeds on `master` with the four-byte signature `0xdeadbeef`. The digest is public via `tn_delegationDigest`, so a permissive handler is enough to bind an arbitrary delegator, and `_getRecipient` sends every later `claimStakeRewards` and `unstake` there.

### Fix

Route on what kind of account the validator actually is rather than on code size:

```solidity
if (_isDelegationDesignator(validatorAddress)) {
    return ECDSA.tryRecoverCalldata(digest, signature) == validatorAddress;
}
return SignatureCheckerLib.isValidSignatureNowCalldata(validatorAddress, digest, signature);
```

The designator is matched by its `0xef0100` prefix, not merely its length. EIP-3541 forbids deployed code from starting with `0xEF`, so the prefix is unambiguous, and a real 23-byte contract still routes to ERC-1271 (`testRevert_delegateStake_twentyThreeByteContractIsNotADesignator`).

Delegated EOAs are now held to their secp256k1 key exactly as undelegated ones are; only genuine contract accounts reach ERC-1271, and `test_delegateStake_contractValidator_usesERC1271` pins that we did not break them. An operator who has moved to a smart account and no longer holds the key still has two routes in: `stake`, which authorizes off `msg.sender`, and the governance path, which skips the signature check entirely.

## #164 - 7702 dissolves EOA screening

The premise holds - screening at mint time is not durable, a holder can attach a reverting delegation afterwards - but **the described exploit chain does not reproduce.** `burn` and the slash-to-zero branch of `applySlashes` both succeed today against a validator whose account rejects every incoming call.

`_consensusBurn` zeroes `balances[validatorAddress]` and consolidates the stake on Issuance *before* reaching `_unstake`, so `_unstake` returns at its zero-balance guard and never calls `distributeStakeReward`. No value is pushed to the validator, and there is no burn payout to convert to a pull-based credit.

What was true is that the property was incidental rather than structural, and actively obscured: `_consensusBurn` called a function named `_unstake`, passing a `recipient` it computed and never used. So rather than a no-op fix, this PR makes the property explicit and pins it:

- `_burnConsensusNFT` is split out of `_unstake` (token burn, supply guard, delegation clear - no value movement) and the confiscating path calls it directly. Behavior is unchanged; the misleading payout call is gone.
- Four regression tests cover both ejection paths against hostile account code: a reverting 7702 delegation on the validator, on the delegator (the actual payout recipient when delegated), and the queued version-change escrow, which is the one balance an ejection owes back and which already credits `claimRefund` rather than pushing. The slash test also concludes the following epoch, since that path runs inside a system call where a revert stalls the closing block rather than merely inconveniencing governance.

These four pass on `master` too. That is the point: they document a property that currently holds by accident and would otherwise be easy to refactor away.

`nonReentrant` is deliberately not added to `burn`. It is `onlyOwner` and, after this change, provably makes no call to an untrusted address.

## Testing

- 11 new tests in `test/consensus/ConsensusRegistryEIP7702Test.t.sol`
- Verified against `master`: the four #163 tests fail (two `NotValidator`, two "did not revert as expected"), the four #164 tests pass
- Full suite green under both `fast` and `ci` profiles: 324 passed, 0 failed, 2 skipped

## Also included

`artifacts/ConsensusRegistry.json` regenerated, since the source changed and the committed artifact is consumed downstream. Docs updated: the `delegateStake` natspec now states which authority is competent to sign, and `design.md` gains bullets for delegation authorization under account abstraction and for the push-free ejection property.


Huwonk:
## Audit pass before review

Re-reviewed the change adversarially. No defects found; three gaps closed in `196c62c` (tests and docs only, no source change).

### Verified

| Check | Result |
|---|---|
| Storage layout vs `master` | **Identical.** Confirmed with `forge inspect storageLayout` on both. The new constant is `constant`, so it lives in bytecode, not a slot. Matters because the registry is genesis-injected and upgraded in place |
| ABI vs `master` | **Identical.** No external signature changed; only bytecode needs to move downstream |
| Runtime bytecode | 34,305 -> 34,699 bytes (+394). Already far past EIP-170 by design (genesis-injected, never CREATEd), so the delta changes nothing |
| Other codesize-based auth in the repo | **None.** `delegateStake` is the only signature-verification site in `src/`; `_hashTypedData` is reached from exactly two places, and no other contract branches on code size |
| `_burnConsensusNFT` refactor | Behaviour-identical. The removed `_unstake` tail was a pure read plus an early return on `bal == 0`, and the `recipient` argument was computed but unused. `_unstake` has no overrides |
| `.code.length` is EXTCODESIZE-only | **Confirmed empirically.** 193 gas against an empty EOA, a tiny contract, and the 34.7KB registry alike. No code-copy cost, no griefing surface from a large-code validator |
| Designator prefix is unambiguous | EIP-3541 forbids deployed code from starting with `0xEF`, so only a 7702 designator can wear it. A 23-byte contract that is not one still routes to ERC-1271 (tested) |
| Delegation revocation | Delegating to `address(0)` clears the account's code entirely, so a revoked account is codesize 0 and takes the plain EOA path |
| Existing delegations | Unaffected. The check runs only at `delegateStake` time and only for `Undefined` validators |
| Reentrancy | The change strictly reduces external-call surface: the 7702 route now makes no call at all where it previously made a full-gas ERC-1271 staticcall |

### Gaps closed

Three cases were uncovered on the delegated-EOA route, now tested and each failing against the pre-fix source:

- 65-byte signature with `v` outside `{27, 28}` -> rejected
- empty signature (recovers `address(0)`) -> rejected
- **route parity**: the designator branch accepts exactly the signature set the plain EOA branch accepts

The parity test probes the one real divergence candidate. Neither solady route screens high-s, so `(r, n-s, v^1)` is accepted by both - verified directly, `true`/`true`. That is pre-existing behaviour, unchanged by this PR, and inert here: the digest binds the delegator and a nonce, and a successful call leaves `Undefined` status behind, so a malleated signature authorizes nothing new and cannot be replayed or front-run. Worth stating explicitly rather than leaving implied, since "held to their key exactly as undelegated EOAs are" is only true if the two routes agree. On `master` this test fails `true != false`.

`invariants.md` now states the two properties to check the code against: which authority `delegateStake` accepts per account kind, and that neither ejection path makes a recipient-facing external call.

### Behaviour change worth a reviewer's eye

An operator who moved to a 7702 smart account and no longer controls the underlying key can no longer be onboarded through `delegateStake`. That is the intended trade - the alternative accepts a revocable wallet program as a stand-in for the validator identity - and two routes remain: `stake`, which authorizes off `msg.sender`, and the governance path, which skips the signature check. Documented in `design.md`.

### Coordination

Bytecode changed, so the node's genesis allocation needs the regenerated `artifacts/ConsensusRegistry.json` picked up on the next submodule bump. No ABI change, so nothing else downstream moves.

### Test status

- 14 tests in `ConsensusRegistryEIP7702Test.t.sol`; 8 of them fail against `master`, 6 (the ejection ones) pass on both by design
- Full suite: 327 passed, 0 failed, 2 skipped under the `ci` profile
- Heavy fuzz pass over all consensus suites at 2000 runs: 227 passed, 0 failed

grantkee (approved):

chasebrownn (approved):

